PR #1, opened 2017-10-02
Perl 5.24 compatibility

davea:
A few packages need upgrading due to `.` not being in `@INC` any more.

(not sure if the entire `cpanfile.snapshot` diff is required, it seems like a lot of stuff got removed?)

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -1,5 +1,5 @@
 # setenv script
-requires 'List::MoreUtils';
+requires 'List::MoreUtils', '>= 0.425';
 requires 'local::lib';
 requires 'Class::Unload';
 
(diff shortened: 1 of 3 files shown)


cpanfile
```diff
@@ -1,15 +1,20 @@
 # setenv script
-requires 'List::MoreUtils';
+requires 'List::MoreUtils', '>= 0.425';
 requires 'local::lib';
 requires 'Class::Unload';
 
+# Specific versions of dependencies
+requires 'B::Hooks::OP::Check', '>= 0.20';
+requires 'XML::NamespaceSupport', '>= 1.12';
```
dracos: Once these two are updated in the snapshot, as they have been, we don't need to keep these lines in the cpanfile.

cpanfile
```diff
@@ -1,15 +1,20 @@
...
+requires 'Module::Install';
```
dracos: As mentioned in chat, if the equivalent of https://github.com/mysociety/fixmystreet/commit/f52a086bce1dee6110aed24f85670d35b92f5ff3 is done in this repo, then I think you can drop the Module::Install installation. This will remove all the new modules from the snapshot.

davea:
Thanks for the review! I've addressed those points and pulled in the updated `carton`, and it deploys cleanly from scratch on stretch now 🎉

PR #6, opened 2018-02-23
Support for Confirm integrations

davea:
This PR adds support for Confirm. I've mostly left the commits as they occurred, with some minor reordering and squashing for brevity.

The branch ~~starts life as a BANES-specific integration, and ends up as~~ is a fairly generic base integration that supports all the features in use right now for BANES and Buckinghamshire's integrations.

PR #7, opened 2018-02-23
BANES Confirm integration

davea:
This PR adds all of the BANES-specific code for their integration. I have left the commits mostly as they occurred so you can see the history of how things progressed and perhaps the rationale behind the structure.

### DIFF
--- a/perllib/Integrations/Confirm/BANES.pm
+++ b/perllib/Integrations/Confirm/BANES.pm
@@ -1,0 +1,18 @@
+package Integrations::Confirm::BANES;
+
+use Path::Tiny;
+use Moo;
+extends 'Integrations::Confirm';
+with 'Role::Config';
+
+has config_filename => (
+    is => 'ro',
+    default => 'banes_confirm',
+);
+
+sub _build_config_file {
+    my $self = shift;
+    path(__FILE__)->parent(4)->realpath->child('conf/council-' . $self->config_filename . '.yml');
+}
+
+1;
--- a/perllib/Open311/Endpoint/Integration/UK/BANES.pm
+++ b/perllib/Open311/Endpoint/Integration/UK/BANES.pm
@@ -1,0 +1,19 @@
+package Open311::Endpoint::Integration::UK::BANES;
+
+use Moo;
+extends 'Open311::Endpoint::Integration::Confirm';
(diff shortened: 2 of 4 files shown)


perllib/Integrations/Confirm/BANES.pm
```diff
@@ -15,4 +15,15 @@ sub credentials {
     );
 }
 
+sub server_timezone { 'Europe/London' }
+
+has '+memcache_namespace'  => (
+    is => 'ro',
```
dracos: No need for this line if you're using the `+`.
(Same applies to all the attributes in this PR.)

perllib/Open311/Endpoint/Integration/UK/BANES.pm
```diff
@@ -5,52 +5,161 @@ use Moo;
 
 use Integrations::Confirm::BANES;
 
-
 has jurisdiction_id => (
     is => 'ro',
     default => 'banes_confirm',
 );
 
-sub integration_class { 'Integrations::Confirm::BANES' }
```
dracos: Guess would be good to not have all this in the parent PR if it's all now changed.

davea: Not sure I follow you here. `integration_class` comes back below as an attribute.

dracos: I just meant I don't really want any BANES-specific stuff in #6 especially if it's then rewritten in this PR :)

davea: Ah I see what you mean. I tried to break the PRs apart at the most logical point, but the commit history has the BANES-specific code fairly intertwined as that's how it progressed. I can revisit it and extract the council-specific stuff out of #6 if you want, but once #6, #7 and #8 are merged I don't think it'd be worth the hassle.

dracos:
This and #8 now build upon config being in an external YAML file as approved in #12. I've kept copies of old branches so we can set up the YAML files in config (or if something is wrong with my rebasing).

dracos:
The merge fails because it doesn't seem to take the change to .travis.yml along, not sure why not. But it should be fine.

PR #10, opened 2018-03-19
Implement get_service_requests to return Enquiries from Confirm

dracos:
Pulled out from now-merged #6.


davea:
👍 Looks good here!

PR #12, opened 2018-03-22
[BANES] Config in YAML file.

dracos:


dracos:
I've pushed something for better testing/discussion. I think it works - it doesn't look like currently you can pass stuff to `integration_class` as SOAP::Lite calls `new()`, so instead have kept Integrations::Confirm::BANES in order to set the 'jurisdiction_id' now needed to load the config file. Which I think makes sense, if not amazing.
I put username/password/tenant_id in as separate, but they could all go under a key as you say if that's preferable.

dracos:
I've taken the Confirm bit out and put on master and rewritten previous PR to match, so this no longe needed.

PR #21, opened 2018-06-06
[Confirm] Retry EnquiryUpdate if rejected for too early LoggedTime

davea:
Stops those annoying "Logged Date 04/06/2018 15:33:28 must be greater than the Effective Date of current status log" errors.

### DIFF
--- a/perllib/Integrations/Confirm.pm
+++ b/perllib/Integrations/Confirm.pm
@@ -138,12 +138,13 @@ sub ProcessOperationsRequest {
 
 sub perform_request {
     my $self = shift;
-
-    my ($username, $password, $tenant) = $self->credentials;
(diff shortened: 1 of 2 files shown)


perllib/Integrations/Confirm.pm
```diff
@@ -138,13 +138,14 @@ sub ProcessOperationsRequest {
 
 sub perform_request {
     my $self = shift;
-
-    my ($username, $password, $tenant) = $self->credentials;
+    my $opts = ref $_[-1] eq 'HASH' ? pop : {};
```
dracos: Nice :)

davea: I had to resist the urge to call it `$kwargs` ;)

PR #23, opened 2018-09-11
Oxfordshire WDM integration

struan:
Adds integration with WDM for OCC.

Only handles new reports, new updates and fetching updates. OCC categories are configured on FMS for the time being.

### DIFF
--- a/perllib/Integrations/WDM.pm
+++ b/perllib/Integrations/WDM.pm
@@ -31,6 +31,11 @@ has 'requests_endpoint' => (
     default => 'requests.xml',
 );
 
+has 'updates_endpoint' => (
+    is => 'ro',
+    default => 'updates.xml',
+);
+
 has service_request_content => (
     is => 'ro',
     default => '/open311/service_request_extended'
@@ -44,6 +49,15 @@ sub format_datetime {
     );
 
     return $fmt->format_datetime($dt)
+}
+
+sub parse_w3c_datetime {
+    my ($self, $dt_string) = @_;
+
+    my $w3c = DateTime::Format::W3CDTF->new;
+    my $dt = $w3c->parse_datetime($dt_string);
+
+    return $dt;
 }
 
 sub post_request {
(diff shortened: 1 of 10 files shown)


perllib/Open311/Endpoint/Integration/WDM.pm
```diff
@@ -0,0 +1,60 @@
+package Open311::Endpoint::Integration::WDM;
+
+use Moo;
+extends 'Open311::Endpoint';
+with 'Open311::Endpoint::Role::mySociety';
+
+use DateTime::Format::Strptime;
+
+use Integrations::WDM;
+use Open311::Endpoint::Service::UKCouncil::Oxfordshire;
```
davea: This creates a coupling between WDM and Oxfordshire-specific stuff. Will that cause issues if/when another WDM integration comes along?

perllib/Open311/Endpoint/Integration/WDM.pm
```diff
@@ -42,6 +53,33 @@ sub post_service_request {
...
+        # ignore non FMS references
+        next unless $service_request_id =~ /^\d+$/;
+        my $customer_reference = $update->{ENQUIRY_REFERENCE};
+        $customer_reference =~ s/\s*$//;
+        push @updates, Open311::Endpoint::Service::Request::Update::mySociety->new(
+            status => lc $update->{STATUS},
+            update_id => $update->{UpdateID},
+            service_request_id => $service_request_id,
+            customer_reference => $customer_reference,
+            description => $update->{COMMENTS},
```
davea: In the Confirm integration we have a `publish_service_update_text` flag which must be set in order for the comment text to be published. Worth doing something similar here, or have WDM users been trained about what will/won't be published?

perllib/Integrations/WDM.pm
```diff
@@ -118,6 +118,43 @@ sub post_request {
...
+            }
+        }
+    };
+    my $response = $self->soap_post($self->updates_endpoint, 'UpdateWdmEnquiry', 'xDoc', $self->_create_xml_string($data));
+
+    my $resp_text = $response->valueof('//UpdateWdmEnquiryResponse/UpdateWdmEnquiryResult');
+    if ($resp_text eq 'OK') {
+        return {
+            status => $args->{status},
+            # XXX I am not sure this is correct
```
davea: Was this resolved, can the comment be removed/updated?

perllib/Open311/Endpoint/Integration/WDM.pm
struan: The latter is the plan.

perllib/Open311/Endpoint/Integration/WDM.pm
struan: I think we'll probably need to see when that happens. It's hard to tell at the moment what's generic about this and what's custom.

PR #25, opened 2018-10-30
[OCC] handle incorrect IDs in old comment updates

struan:
When fetching comments the serice_request_id and the customer_reference
are the wrong way round so swap them over if it looks like an old
request. And do not pass a customer_reference because we don't want to
set that on old enquiries initially.

### DIFF
--- a/perllib/Open311/Endpoint/Integration/WDM.pm
+++ b/perllib/Open311/Endpoint/Integration/WDM.pm
@@ -87,7 +87,8 @@ sub get_service_request_updates {
         $customer_reference =~ s/\s*$//;
         my $status = lc $update->{STATUS};
         $status =~ s/ /_/g;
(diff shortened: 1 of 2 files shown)


perllib/Open311/Endpoint/Integration/WDM.pm
```diff
@@ -87,14 +87,22 @@ sub get_service_request_updates {
...
+        my %args = (
             status => $status,
             update_id => $update->{UpdateID},
             service_request_id => $service_request_id,
             customer_reference => $customer_reference,
             description => $update->{COMMENTS},
             updated_datetime => $self->parse_w3c_datetime($update->{UPDATE_TIME}),
         );
+
+        if ( $customer_reference !~ /^ENQ/ ) {
```
davea: Could you add a comment explaining why this is here? If/when we add another WDM integration it'd be useful to know that this is specific to a particular installation.

PR #30, opened 2019-01-08
add logging

struan:


cpanfile
```diff
@@ -41,3 +42,4 @@ requires 'Test::LongString';
 requires 'Test::MockModule';
 requires 'Test::MockTime';
 requires 'Test::More';
+requires 'File::Slurp';
```
dracos: Path::Tiny already has better slurp stuff built in, so would rather use that than add File::Slurp (also adding it appears to have removed some things from the snapshot).

perllib/Open311/Endpoint/Logger.pm
```diff
@@ -0,0 +1,84 @@
...
+
+has logger => (
+    is => 'lazy',
+    default => sub {
+        my $self = shift;
+        return undef unless $self->config->{logfile};
+        my $min_level = $self->config->{min_log_level} || 'error';
+        Log::Dispatch->new(
+            outputs => [
+                [ 'File',
```
dracos: Should this be File::Locked?

dracos: Also, perhaps we could use https://metacpan.org/pod/Log::Dispatch::File::Stamped ? Depends how much logging we're going to be doing :)

struan: Probably Locked, yes.

I talked to sam and he said as long as we stick it in the applogs directory it will be rotated and that we shouldn't be using timestamped file names.

PR #33, opened 2019-04-03
[Bexley] Initial update fetching code.

dracos:
Fetches CSV file of updates, filters to within the requested date range, and outputs matches as Open311 updates for FMS to consume.

### DIFF
--- a/perllib/Open311/Endpoint/Integration/Symology.pm
+++ b/perllib/Open311/Endpoint/Integration/Symology.pm
@@ -3,9 +3,13 @@ package Open311::Endpoint::Integration::Symology;
 use v5.14;
 use warnings;
 
+use DateTime::Format::Strptime;
+use DateTime::Format::W3CDTF;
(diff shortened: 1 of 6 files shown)


perllib/Open311/Endpoint/Integration/Symology.pm
```diff
@@ -230,10 +246,51 @@ sub check_error {
...
+
+    my @updates;
+    while ($csv->getline($fh)) {
+        my $dt = $self->date_formatter->parse_datetime($row->{date_history});
+        next unless $dt >= $start_time && $dt <= $end_time;
+
+        my $status;
+        $status = 'investigating' if $row->{stage} == 1 && $row->{action_due};
+        $status = 'fixed' if $row->{stage} == 9 && !$row->{action_due};
+        $status = 'no_further_action' if $row->{stage} == 9 && $row->{lca} eq 'ATS';
```
struan: Should we not have a default status here or something that handles one of these conditions not being true?

PR #36, opened 2019-04-22
[Rutland] ignore reports older than start date

struan:
Because we don't have a way to limit requests by date ignore those that
are older than the start date as they will only be rejected by
FixMyStreet anyway. Also means we avoid the calls to fetch service
details for those requests.

### DIFF
--- a/perllib/Open311/Endpoint/Integration/SalesForce.pm
+++ b/perllib/Open311/Endpoint/Integration/SalesForce.pm
@@ -128,6 +128,10 @@ sub get_service_requests {
     my $w3c = DateTime::Format::W3CDTF->new;
     my @updates = ();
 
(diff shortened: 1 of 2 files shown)


perllib/Open311/Endpoint/Integration/SalesForce.pm
```diff
@@ -137,6 +137,11 @@ sub get_service_requests {
 
         my $update_time = $self->parse_datetime($request->{LastModifiedDate});
         my $request_time = $self->parse_datetime($request->{requested_datetime__c}) || $update_time;
+
+        if ($args->{start_date}) {
+            my $start_time = DateTime::Format::W3CDTF->parse_datetime($args->{start_date});
```
dracos: This doesn't change each iteration, so could be done once at the start.

PR #44, opened 2019-06-06
[Confirm] allow categories to exist in multiple groups

struan:
Allow categories to be under more than one group in the whitelist. Only
one services entry per category is generated but it will include all the
groups in the group tag.

### DIFF
--- a/perllib/Open311/Endpoint/Integration/Confirm.pm
+++ b/perllib/Open311/Endpoint/Integration/Confirm.pm
@@ -13,6 +13,7 @@ use Open311::Endpoint::Service::Request::CanBeNonPublic;
 
 use Path::Tiny;
 use SOAP::Lite; # +trace => [ qw/method debug/ ];
+use Text::CSV;
 
 
 around BUILDARGS => sub {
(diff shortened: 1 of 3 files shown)


perllib/Open311/Endpoint/Integration/Confirm.pm
```diff
@@ -522,7 +523,9 @@ sub _services {
         }
     }
 
+    my $csv = Text::CSV->new();
     my @services = ();
+    my %service_codes;
     for my $group (keys %{ $self->service_whitelist }) {
```
dracos: Probably want to stick a sort in here, so that the ordering of multiple groups is consistent (which I think is the test failure).

perllib/Open311/Endpoint/Integration/Confirm.pm
```diff
@@ -532,20 +535,29 @@ sub _services {
...
+                    service_code => $code,
+                    description => $name,
+                    group => [$group],
+                    keywords => $private_services{$code} ? [qw/ private /] : [],
+                };
             }
-            push @services, $o311_service;
         }
     }
+    for my $service (values %service_codes) {
```
dracos: Also a sort here for consistent output?

PR #48, opened 2019-06-13
[Alloy] use startDate for updatetime

struan:
Use the version startDate for update times which is the correct thing to do according to Yotta.


t/open311/endpoint/alloy.t
```diff
@@ -97,6 +97,8 @@ $integration->mock('api_call', sub {
             if ( $type =~ /DEFECT/ ) {
                 if ( $time =~ /2019-01-02/ ) {
                     $content = path(__FILE__)->sibling('json/alloy/defect_search_all.json')->slurp;
+                } elsif ( $time =~ /2019-01-03/ ) {
+                    $content = path(__FILE__)->sibling('json/alloy/defect_search_multiple.json')->slurp;
```
davea: This file seems to be missing so the tests aren't happy

struan: oops, sorted now.

PR #50, opened 2019-07-18
[Northamptonshire] always set external_status_code on defect updates

struan:
We need to set this to stop phantom updates being produced. This
happens because when an inspection is closed it always sets an
external_status_code which we never unset. Then when updates arrive
from defects with no external_status_code the template fetching code at
FixMyStreet sees that the external_status_code has changed and fetches
the template. This means we always get an update even if nothing has
changed. So, set the external_status_code on defects associated with an
inspection to the external_status_code used when an inspection is marked
for raising as a defect.

### DIFF
--- a/perllib/Open311/Endpoint/Integration/Alloy.pm
+++ b/perllib/Open311/Endpoint/Integration/Alloy.pm
@@ -318,6 +318,8 @@ sub get_service_request_updates {
     }
 
     # updates to defects
+    my $closure_mapping = $self->config->{inspection_closure_mapping};
+    my %reverse_closure_mapping = map { $closure_mapping->{$_} => $_ } keys %{$closure_mapping};
     $updates = $self->fetch_updated_resources($self->config->{defect_resource_name}, $args->{start_date});
     for my $update (@$updates) {
         next if $self->is_ignored_category( $update );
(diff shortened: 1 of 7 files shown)


perllib/Open311/Endpoint/Integration/Alloy.pm
```diff
@@ -318,6 +318,8 @@ sub get_service_request_updates {
     }
 
     # updates to defects
+    my $closure_mapping = $self->config->{inspection_closure_mapping};
+    my %reverse_closure_mapping = map { $closure_mapping->{$_} => $_ } keys %{$closure_mapping};
```
dracos: I realise only action scheduled is then used, don't know if worth noting that this isn't 1-1 so some reverse values will be lost.

perllib/Open311/Endpoint/Integration/Alloy.pm
```diff
@@ -375,6 +377,14 @@ sub get_service_request_updates {
...
         );
 
+        # we need to set this to stop phantom updates being produced. This happens because
+        # when an inspection is closed it always sets an external_status_code which we never
+        # unset. Then when updates arrive from defects with no external_status_code the template
+        # fetching code at FixMyStreet sees that the external_status_code has changed and fetches
+        # the template. This means we always get an update even if nothing has changed. So, set
+        # this to the external_status_code used when an inspection is marked for raising as a
+        # defect.
+        $args{external_status_code} = $reverse_closure_mapping{'action_scheduled'} if $fms_id or $linked_defect;
```
dracos: I think this means that even when the defect returns a 'fixed' state, say, this will still pass the action scheduled external status code through as well, which means no template will match at the FMS end.

PR #60, opened 2019-09-24
[Bexley] Initial Confirm boilerplate, and multiple children controller

dracos:
This PR moves the Bexley Symology class to be Bexley::Symology, creates an initial Confirm setup as Bexley::Confirm, and then creates a new Bexley class that combines those two together as a united front to be exposed - joining services together, sending new reports to the correct one of the two, and so on.


perllib/Open311/Endpoint/Integration/Confirm.pm
```diff
@@ -524,23 +524,26 @@ sub _services {
 
     my @services = ();
     my %service_codes;
-    for my $group (sort keys %{ $self->service_whitelist }) {
-        my $whitelist = $self->service_whitelist->{$group};
+    my $fetch_all_services = 0;
```
struan: How does this get changed?

dracos: It was more for local testing of a Confirm instance, so I had manually changed it to 1, run it locally to see all the services, then changed it back. Other suggestions welcome :)

dracos:
I have rebased this to cut out the initial Confirm commit (moved to bexley-confirm branch), so this only tests the existing Symology, plus adds a "default non-prefixed subclass" - Symology - which will then not have the prefix added that other subclasses will. This means nothing needs to be changed at the FixMyStreet end when this code goes into effect, and taking out the Confirm bit means it can go into effect whenever (now), rather than waiting for Confirm data to be ready.

PR #61, opened 2019-09-25
[Alloy] change the version id we use for updates numbers

struan:
use `usedSystemVersionId` instead of `resourceSystemVersionId`.

### DIFF
--- a/perllib/Open311/Endpoint/Integration/Alloy.pm
+++ b/perllib/Open311/Endpoint/Integration/Alloy.pm
@@ -329,7 +329,7 @@ sub get_service_request_updates {
             my %args = (
                 status => $status,
                 external_status_code => $reason_for_closure,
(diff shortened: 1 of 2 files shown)


perllib/Open311/Endpoint/Integration/Alloy.pm
```diff
@@ -329,7 +329,7 @@ sub get_service_request_updates {
             my %args = (
                 status => $status,
                 external_status_code => $reason_for_closure,
-                update_id => $resource->{version}->{resourceSystemVersionId},
+                update_id => $resource->{version}->{usedSystemVersionId},
```
dracos: Does this also apply to line 396?
(I mean, I guess it doesn't have to as that is only looking at current state when these two IDs look to be equivalent.)

PR #62, opened 2019-09-26
[Confirm] Output <groups> correctly for wrapped_services

davea:
Ensures any existing groups in `wrapped_services` in the YML config are included in the output. Also allows a wrapped service to appear in multiple groups by setting the `group` value to a list instead of a string.

### DIFF
--- a/perllib/Open311/Endpoint/Integration/Confirm.pm
+++ b/perllib/Open311/Endpoint/Integration/Confirm.pm
@@ -728,7 +728,9 @@ sub _wrap_services {
     for my $code (keys %{$self->wrapped_services}) {
         if ($self->wrapped_services->{$code}->{passthrough}) {
             my $original_service = $original_services{$code};
-            $original_service->group($self->wrapped_services->{$code}->{group} || $original_service->group);
+            my $wrapped_group = $self->wrapped_services->{$code}->{group};
+            $wrapped_group = [$wrapped_group] if ( $wrapped_group && ref $wrapped_group ne 'ARRAY' );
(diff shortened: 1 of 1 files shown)


perllib/Open311/Endpoint/Integration/Confirm.pm
```diff
@@ -756,11 +758,13 @@ sub _wrap_services {
             );
         }
 
+        my $groups = $self->wrapped_services->{$code}->{group};
+        $groups = [$groups] if ref $groups ne 'ARRAY';
```
dracos: `if $groups && ref $groups` to be consistent with the above (even if not necessary?)

perllib/Open311/Endpoint/Integration/Confirm.pm
```diff
@@ -728,7 +728,9 @@ sub _wrap_services {
     for my $code (keys %{$self->wrapped_services}) {
         if ($self->wrapped_services->{$code}->{passthrough}) {
             my $original_service = $original_services{$code};
-            $original_service->group($self->wrapped_services->{$code}->{group} || $original_service->group);
+            my $wrapped_group = $self->wrapped_services->{$code}->{group};
+            $wrapped_group = [$wrapped_group] if ( $wrapped_group && ref $wrapped_group ne 'ARRAY' );
+            $original_service->groups($wrapped_group || $original_service->group);
```
dracos: whoops, this should be groups at the end!

dracos:
Did the above, merged/deployed, Lincs is okay, looks like.

PR #63, opened 2019-09-26
[Bexley Confirm] Add test for getting the list of services

chrismytton:
This tests that the list of services is correctly returned from the Bexley Confirm adapter.

This has mostly been copied from `t/open311/endpoint/confirm.t` and then slimmed down to just have the bits to test the GET services call. I must admit that the `$open311->mock(perform_request => sub {…` part is a bit of a mystery to me, for now I've just copy/pasted it in to get things working, so there's a fair bit of duplication at the moment.

Also the `*::Dummy` packages seem to be performing overlapping duties as they're both setting the config file name, but again I've just copied them in and adjusted them to be Bexley-Confirm-specific for now.

Any and all feedback welcome!

## User story

As a council I want users to be able to select from a list of problem categories.

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1568

### DIFF
--- a/t/open311/endpoint/bexley_confirm.t
+++ b/t/open311/endpoint/bexley_confirm.t
@@ -25,12 +25,7 @@ use strict;
 use warnings;
 
 use Test::More;
-use Test::LongString;
 use Test::MockModule;
-use Test::Output;
-
-use JSON::MaybeXS;
-use Path::Tiny;
 
 BEGIN { $ENV{TEST_MODE} = 1; }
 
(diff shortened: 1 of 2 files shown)


t/open311/endpoint/bexley_confirm.t
```diff
@@ -0,0 +1,119 @@
...
+            { EnquiryNumber => 2002, EnquiryStatusLog => [ { EnquiryLogNumber => 2, LogEffectiveTime => '2018-01-17T12:34:56Z', LoggedTime => '2018-03-01T13:30:00.4000Z', EnquiryStatusCode => 'DUP' } ] },
+        ] } } };
+    }
+    return {};
+});
+
+use Open311::Endpoint::Integration::UK::Bexley::Confirm::Dummy;
+
+my $endpoint = Open311::Endpoint::Integration::UK::Bexley::Confirm::Dummy->new(
+    jurisdiction_id => 'dummy',
```
dracos: I realise this is copied, but looks like confirm.t is wrong too - these two lines (91/92) aren't needed because they're both then manually set by the BUILDARGS at line 15/16, I think (though see my other confused comment that I'll try and work out now).

t/open311/endpoint/bexley_confirm.t
```diff
@@ -0,0 +1,119 @@
...
+sub _build_config_file { path(__FILE__)->sibling("bexley_confirm.yml")->stringify }
+
+package Open311::Endpoint::Integration::UK::Bexley::Confirm::Dummy;
+use Path::Tiny;
+use Moo;
+extends 'Open311::Endpoint::Integration::UK::Bexley::Confirm';
+around BUILDARGS => sub {
+    my ($orig, $class, %args) = @_;
+    $args{jurisdiction_id} = 'dummy';
+    $args{config_file} = path(__FILE__)->sibling("bexley_confirm.yml")->stringify;
```
dracos: On the two places, yes, this is because there are two places that read the same configuration file (technically, I guess they could read different configuration files, but never have so far) - one is the actual integration code for talking to the backend, the other is the Open311-integration "mapping" business code, as it were.

t/open311/endpoint/bexley_confirm.t
```diff
@@ -0,0 +1,119 @@
...
+use Test::MockModule;
+use Test::Output;
+
+use JSON::MaybeXS;
+use Path::Tiny;
+
+BEGIN { $ENV{TEST_MODE} = 1; }
+
+my $open311 = Test::MockModule->new('Integrations::Confirm::Bexley');
+$open311->mock(perform_request => sub {
```
dracos: Okay, so what this is doing is saying that if any code ever calls the perform_request function in Integrations::Confirm::Bexley, then call this function here instead. You can see that function defined in Confirm.pm, it is the actual function that makes a call out to a Confirm server (by calling ProcessOperationsRequest). So this function intercepts all such calls, checking that what is passed is okay and returning what would make sense for a Confirm server to return.
For the purposes of the test you have, therefore, you'd only need the "GetEnquiryLookups" bit, because that's the Confirm call that's fetching the services from Confirm.

t/open311/endpoint/bexley_confirm.yml
```diff
@@ -0,0 +1,15 @@
+endpoint_url: "http://example.org/endpoint"
+username: "username"
+password: "pw"
+tenant_id: "123"
+server_timezone: Europe/London
+default_site_code: 999999
+service_whitelist:
+  Flooding & Drainage:
```
dracos: No harm in checking a category in multiple groups (which is what this is doing), but I don't think Bexley will be using this, so might be more straightforward to have a couple of different categories instead.

t/open311/endpoint/bexley_confirm.t
```diff
@@ -0,0 +1,119 @@
...
+use Moo;
+extends 'Open311::Endpoint::Integration::UK::Bexley::Confirm';
+around BUILDARGS => sub {
+    my ($orig, $class, %args) = @_;
+    $args{jurisdiction_id} = 'dummy';
+    $args{config_file} = path(__FILE__)->sibling("bexley_confirm.yml")->stringify;
+    return $class->$orig(%args);
+};
+has integration_class => (is => 'ro', default => 'Integrations::Confirm::Bexley::Dummy');
+sub jurisdiction_id { return 'dummy'; }
```
dracos: I am a bit confused as to why this line is needed (given the BUILDARGS above seems like it should be setting it) - something weird going on here I don't yet understand.

t/open311/endpoint/bexley_confirm.t
```diff
@@ -0,0 +1,119 @@
...
+
+package main;
+
+use strict;
+use warnings;
+
+use Test::More;
+use Test::LongString;
+use Test::MockModule;
+use Test::Output;
```
dracos: Not currently using this, or Test::LongString (but you might be as you add more, I assume).

t/open311/endpoint/bexley_confirm.t
```diff
@@ -0,0 +1,119 @@
...
+
+use JSON::MaybeXS;
```
dracos: Not using any JSON decoding at present either, or any Path::Tiny that I can see (can't see confirm.t using any Path::Tiny either).

t/open311/endpoint/bexley_confirm.t
dracos: I've fixed this in #64 which will then mean this line can be removed.

PR #66, opened 2019-09-30
[Bexley Confirm] Test getting existing reports

chrismytton:
This tests that existing enquiries are returned from the Bexley Confirm adapter.

Note that I rewrote the previous `bexley_confirm.t` changes to simplify them, see 60ad0d47b152a6fbf1d5c6859073024e1ba7262c. That's why this change is adding the `::Dummy` classes.

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1569


chrismytton:
@dracos Thanks for the review! All makes sense. I'll add some tests to `confirm.t` ensure the user information is removed, then look at the FMS side.

PR #68, opened 2019-10-04
[Bexley Confirm] Split into two separate integrations - trees and grounds

chrismytton:
Bexley have two separate Confirm servers, one for trees and one for grounds. This change splits the Bexley Confirm integration in two so we can have one integration for each server.

There aren't separate tests for the trees and grounds integrations, since the implementations are virtually identical. I spoke to @dracos about this and we agreed that if/when the implementations diverge we can add tests at that point.

Related to https://github.com/mysociety/fixmystreet-commercial/issues/1568 and https://github.com/mysociety/fixmystreet-commercial/issues/1573 (since we need separate category and status mappings for the trees and grounds servers).

### DIFF
--- a/t/open311/endpoint/bexley.t
+++ b/t/open311/endpoint/bexley.t
@@ -16,15 +16,6 @@ my $confirm_grounds = Test::MockModule->new('Open311::Endpoint::Integration::UK:
 $confirm_grounds->mock(services => sub {
     return ( new_service('A_BC'), new_service('D_EF') );
 });
-$confirm_grounds->mock(post_service_request_update => sub {
-    my ($self, $args) = @_;
-    is $args->{service_code}, 'D_EF';
-    is $args->{service_request_id}, 1001;
-    return Open311::Endpoint::Service::Request::Update::mySociety->new(
-        status => 'in_progress',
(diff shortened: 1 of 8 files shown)


t/open311/endpoint/bexley.t
```diff
@@ -12,11 +12,24 @@ sub new_service {
...
+    is $args->{service_code}, 'D_EF';
+    is $args->{service_request_id}, 1001;
+    return Open311::Endpoint::Service::Request::Update::mySociety->new(
+        status => 'in_progress',
+        update_id => 456,
+    );
+});
+my $confirm_trees = Test::MockModule->new('Open311::Endpoint::Integration::UK::Bexley::ConfirmTrees');
+$confirm_trees->mock(services => sub {
+    return ( new_service('A_BC'), new_service('D_EF') );
```
dracos: I'd perhaps have one name the same as Grounds, one name different - just to check it wasn't using the wrong service list somehow, say.

chrismytton: Ah yes, good idea! Done in 2b1a34b2240f5aac23d99d1353fbe49b1b5cad7f

PR #71, opened 2019-10-09
[Confirm] Allow override of attribute option names.

dracos:
One service we connect to sends through option text with number prefixes, which we wish to remove before display.


t/open311/endpoint/confirm_attributes.t
```diff
@@ -0,0 +1,223 @@
...
+
+package main;
+
+use Test::More;
+use Test::LongString;
+use Test::MockModule;
+
+BEGIN { $ENV{TEST_MODE} = 1; }
+
+my ($IC, $SIC, $DC);
```
davea: These don't seem to be used anywhere.

PR #74, opened 2019-10-23
[Confirm] Test that StatusLogNotes doesn't appear in updates

chrismytton:
We don't want to leak the StatusLogNotes that we send to Confirm in the open311-adapter output.

This came up when discussing https://github.com/mysociety/fixmystreet-commercial/issues/1574 with @dracos.


chrismytton:
@dracos Turns out it _still_ wasn't actually testing the correct thing, because I'd got the timestamps wrong in the conditional (needed to account for daylight savings). I've pushed 7d83f86 to fix that and add the extra check you suggested to ensure we're getting back the update we're expecting.

PR #81, opened 2019-12-19
Extract multiple children integration from Bexley

chrismytton:
We want to reuse this code for Peterborough, so extract it from Bexley and create a generic Multi backend.

This allows a council to have more than one backend, and handles calling them all and combining the results, as well as sending reports to the correct backend.

Part of the Peterborough ezytreev integration work.

### DIFF
--- a/perllib/Open311/Endpoint/Integration/UK/Bexley.pm
+++ b/perllib/Open311/Endpoint/Integration/UK/Bexley.pm
@@ -1,170 +1,20 @@
 package Open311::Endpoint::Integration::UK::Bexley;
 
 use Moo;
-extends 'Open311::Endpoint';
-with 'Open311::Endpoint::Role::mySociety';
+extends 'Open311::Endpoint::Integration::Multi';
 
 use Module::Pluggable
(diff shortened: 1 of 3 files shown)


perllib/Open311/Endpoint/Integration/UK/Bexley.pm
```diff
@@ -1,170 +1,20 @@
 package Open311::Endpoint::Integration::UK::Bexley;
 
 use Moo;
-extends 'Open311::Endpoint';
-with 'Open311::Endpoint::Role::mySociety';
+extends 'Open311::Endpoint::Integration::Multi';
 
 use Module::Pluggable
     search_path => ['Open311::Endpoint::Integration::UK::Bexley'],
     instantiate => 'new';
```
chrismytton: Would be nice to be able to avoid having to include this in every subclass, but couldn't work out a way to do this in the parent class.

struan: I think it's not that much boiler plate and I suspect anything that avoids this is going to require some slightly hairy code to get round execution order fun.

PR #84, opened 2019-12-20
[Alloy] use correct service_code for fetched reports

struan:
Add the group name to the front of the service code that is included in
fetched defects to match the service codes returned by the service call.
Alongside this add a check to make sure that the defect is using a
supported service code, and if not warn and skip the defect.


struan:
Meged in b5648d8

PR #83, opened 2019-12-20
[Peterborough] Ezytreev integration

chrismytton:
The Peterborough-specific bits of the ezytreev integration.

## Notes

- Follows on from https://github.com/mysociety/open311-adapter/pull/85, that PR should be merged first.

## Before merging

- [ ] Remove commit disabling Confirm integration (the staging version isn't working, so have temporarily disabled it)

PR #85, opened 2020-01-13
[Ezytreev] Add integration

chrismytton:
Integration with the new ezytreev API, which will initially be implemented for Peterborough.

Note that because this is being done for Peterborough there may be some Peterborough-specific assumptions that have unintentionally snuck in here.

The Peterborough-specific bits will be added in a separate PR (currently a WIP: https://github.com/mysociety/open311-adapter/pull/83).

### DIFF
--- a/perllib/Integrations/Ezytreev.pm
+++ b/perllib/Integrations/Ezytreev.pm
@@ -1,0 +1,66 @@
+package Integrations::Ezytreev;
+
+use LWP::UserAgent;
+use HTTP::Request::Common;
+use JSON::MaybeXS;
+use MIME::Base64 qw(encode_base64);
+
+use Moo;
+with 'Role::Config';
+with 'Role::Logger';
+
+has ua => (
+    is => 'lazy',
+    default => sub {
+        my $ua = LWP::UserAgent->new(agent => "FixMyStreet/open311-adapter");
+        my $uri = URI->new($_[0]->config->{endpoint_url});
+        my $netloc = $uri->host . ":" . $uri->port;
+        $ua->credentials($netloc, $uri->host, $_[0]->config->{username}, $_[0]->config->{password});
+        return $ua;
+    },
+);
+
+# Despite the name, this is also used for creating enquiries.
+sub update_enquiry {
+    my ($self, $body) = @_;
+
+    my $url = $self->config->{endpoint_url} . "UpdateEnquiry";
+    my $request = POST $url,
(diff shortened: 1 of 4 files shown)


perllib/Open311/Endpoint/Integration/Ezytreev.pm
```diff
@@ -0,0 +1,190 @@
...
+    my $services = $self->category_mapping;
+    my @services = map {
+        my $name = $services->{$_}{name};
+        my $service = Open311::Endpoint::Service::UKCouncil::Ezytreev->new(
+            service_name => $name,
+            service_code => $_,
+            description => $name,
+            $services->{$_}{group} ? (group => $services->{$_}{group}) : (),
+      );
+    } keys %$services;
```
dracos: I would sort here so that the services are returned in the same order each time (shouldn't matter but might as well).

perllib/Integrations/Ezytreev.pm
```diff
@@ -0,0 +1,68 @@
...
+    my $body = {
+        CRMXRef => $args->{crm_xref},
+        FileName => $photo->filename,
+        Description => "Photo from problem reporter.",
+        FileBase64 => encode_base64($photo->content),
+    };
+
+    my $url = $self->config->{endpoint_url} . "UploadEnquiryDocumentBase64";
+    my $request = POST $url,
+        'Content-Type' => 'application/json',
```
dracos: Just to note you can specify the key here as `Content_Type` if you wish to avoid the quotes.

perllib/Integrations/Ezytreev.pm
```diff
@@ -0,0 +1,68 @@
...
+use MIME::Base64 qw(encode_base64);
+
+use Moo;
+with 'Role::Config';
+with 'Role::Logger';
+
+has ua => (
+    is => 'lazy',
+    default => sub {
+        LWP::UserAgent->new(agent => "FixMyStreet/open311-adapter")
```
dracos: I think at this point, given it's lazy, you could work out the host from the endpoint_url, and call `$ua->credentials(host, realm, $_[0]->config->{username}, $_[0]->config->{password});` and then not have to set it for each call.

chrismytton:
Thanks @dracos! Have pushed a few more commits to address your comments. Will wait for Travis to re-run the tests then get this merged, which will probably tomorrow now.

PR #91, opened 2020-01-29
[Ezytreev] Get update notes for certain status codes

chrismytton:
We want to pass the update notes for certain status codes through to FMS. This change adds a whitelist of codes that we allow the notes for and then checks that list when retrieving updates and shows the notes if it's in the whitelist, otherwise it just shows the enquiry status description.

Fixes https://github.com/mysociety/fixmystreet-commercial/issues/1752

### DIFF
--- a/perllib/Open311/Endpoint/Integration/Ezytreev.pm
+++ b/perllib/Open311/Endpoint/Integration/Ezytreev.pm
@@ -133,16 +133,18 @@ sub get_service_request_updates {
         foreach my $enquiry_status (@{$enquiry->{StatusHistory}}) {
             # Ignore updates created by FMS
             next if $enquiry_status->{StatusByName} eq 'CRM System';
-            my $status = $self->reverse_status_mapping->{$enquiry_status->{EnquiryStatusCode}};
+
+            my $status_code = $enquiry_status->{EnquiryStatusCode};
+            my $status = $self->reverse_status_mapping->{$status_code};
             if (!$status) {
                 $self->logger->warn("Missing reverse status mapping for EnquiryStatus Code " .
-                    "$enquiry_status->{EnquiryStatusCode} (EnquiryStatusID $enquiry_status->{EnquiryStatusID})\n");
+                    "$status_code (EnquiryStatusID $enquiry_status->{EnquiryStatusID})\n");
                 $status = "open";
             }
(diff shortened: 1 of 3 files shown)


perllib/Open311/Endpoint/Integration/Ezytreev.pm
```diff
@@ -133,13 +138,21 @@ sub get_service_request_updates {
                 $status = "open";
             }
             my $dt = $w3c->parse_datetime($enquiry_status->{StatusDate});
-            my $status_description = $enquiry_status->{EnquiryStatusDescription};
-            $status_description =~ s/^\s+|\s+$//g;
+
+            my $description;
+            if (grep { $_ eq $enquiry_status->{EnquiryStatusCode} } @{$self->statuses_to_show_notes_for}) {
```
dracos: Do we also want to check that StatusInfo has some content, and only doing this if yes? (Assuming EnquiryStatusDescription always exists.)

perllib/Open311/Endpoint/Integration/Ezytreev.pm
```diff
@@ -52,6 +52,11 @@ has reverse_status_mapping => (
     default => sub { $_[0]->endpoint_config->{reverse_status_mapping} }
 );
 
+has statuses_to_show_notes_for => (
+    is => 'lazy',
+    default => sub { $_[0]->endpoint_config->{statuses_to_show_notes_for} }
```
dracos: Just idle thought, this could create a hash, e.g. `sub { { map { $_ => 1 } @{$_[0]->endpoint_config->{statuses_to_show_notes_for}} } }` and then the check code would be `$self->statuses_to_show_notes_for->{$enquiry_status->{EnquiryStatusCode}}` (or even nicer if you stick `$enquiry_status->{EnquiryStatusCode}` in a simple var first).

perllib/Open311/Endpoint/Integration/Ezytreev.pm
chrismytton: I think it's fine to always show `StatusInfo`, even if it's empty. Peterborough know that what they type in ezytreev for these status codes will be shown on FMS, so I think it's expected behaviour that when they write nothing then nothing is shown on FMS.

PR #95, opened 2020-02-11
[Oxfordshire] ignore new feature_id attribute

struan:
small change to handle the feature_id asset we're now using now that we have assets.


dracos:
Think it should be a hidden field, not a server set. Not that it matters, as Oxfordshire do not pull services :)

PR #98, opened 2020-02-24
[WDM] enable statuses to be mapped

struan:
add a status_map key to the config and use that to map from WDM statuses
in fetched updates to FixMyStreet statuses


perllib/Integrations/WDM.pm
```diff
@@ -164,6 +164,14 @@ sub get_updates {
     return $updates;
 }
 
+sub map_statuses {
```
davea: This would be better named `map_status` as it only operates on a single status.

PR #105, opened 2020-04-23
[EastSussex] allow groups to be renamed for display

struan:
Some of the internal Salesforce names aren't especially user facing so
allow them to be renamed on the way out and then reversed on the way
back in

### DIFF
--- a/perllib/Open311/Endpoint/Integration/SalesForceRest.pm
+++ b/perllib/Open311/Endpoint/Integration/SalesForceRest.pm
@@ -73,8 +73,10 @@ sub post_service_request {
 
     die "Failed to get user id" unless $user_id;
 
-    $req->{ $mapping->{title} } = $args->{attributes}->{group} || $service->groups->[0];
-    $req->{ $mapping->{group} } = $args->{attributes}->{group} || $service->groups->[0];
+    my $group = $self->_revert_group_name( $args->{attributes}->{group} || $service->groups->[0] );
+
+    $req->{ $mapping->{title} } = $group;
+    $req->{ $mapping->{group} } = $group;
     $req->{ $mapping->{account} } = $user_id;
 
     # most categories use a type and a sub type which map to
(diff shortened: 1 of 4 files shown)


perllib/Open311/Endpoint/Integration/SalesForceRest.pm
```diff
@@ -170,6 +172,31 @@ sub service {
...
+    ];
+
+    return $groups;
+}
+
+sub _revert_group_name {
+    my ($self, $group) = @_;
+
+    my $map = $self->get_integration->config->{group_name_map};
+    my $reverse_map = { map { $map->{$_} => $_ } keys %{ $map } };
```
dracos, line 195: Worth a note the values as well as the keys have to be unique for this to work I guess.

dracos: Note that `my $reverse_map = { reverse %$map }` works :) (could just be a hash rather than a ref too)

perllib/Open311/Endpoint/Integration/SalesForceRest.pm
```diff
@@ -170,6 +172,31 @@ sub service {
...
+        } @{ $groups }
+    ];
+
+    return $groups;
+}
+
+sub _revert_group_name {
+    my ($self, $group) = @_;
+
+    my $map = $self->get_integration->config->{group_name_map};
```
dracos, line 194: Does this need a $map exists check like _rename_groups?

perllib/Open311/Endpoint/Integration/SalesForceRest.pm
```diff
@@ -170,6 +172,31 @@ sub service {
...
+    my $reverse_map = { map { $map->{$_} => $_ } keys %{ $map } };
+
+    return $reverse_map->{$group} ? $reverse_map->{$group} : $group;
```
dracos, line 197: `$reverse_map->{$group} || $group` works here, and in the other function too

perllib/Open311/Endpoint/Integration/SalesForceRest.pm
struan: Given this is for renaming top level categories it'd be a bit odd to two renamed to the same thing so I think that limitation is fine.

PR #111, opened 2020-04-27
[EastSussex] add closest address attribute and use it

struan:
Strip out the extra text we put in and just leave the address as it's
not required in this case.


perllib/Open311/Endpoint/Integration/SalesForceRest.pm
```diff
@@ -258,6 +260,16 @@ sub _get_user {
     return $obj;
 }
 
+sub _add_closest_address {
+    my ($self, $req, $args, $mapping) = @_;
+
+    my $address = $args->{attributes}->{closest_address};
+
+    $address =~ s/^Nearest[^:]*: (.+)[\n\r]*.*/$1/m;
```
dracos, line 268: Not sure I understand the regex - the string that comes in is "Nearest...: ...\n\nNearest...: ...\n\n". So this finds the first line that starts Nearest, and replaces that line, and the next line after newlines with the data from that line. It'll end up with newlines at the end, is that wanted? Plus any text that might be before or after those two lines. If you only want the first Nearest data, perhaps a `m//` to match it and then replace the whole string with that?

PR #120, opened 2020-06-18
[Hackney] Small changes to support defect fetching

davea:
A couple of bug fixes and some customisation for the way Hackney defects are configured in Alloy.

### DIFF
--- a/perllib/Open311/Endpoint/Integration/AlloyV2.pm
+++ b/perllib/Open311/Endpoint/Integration/AlloyV2.pm
@@ -716,7 +716,7 @@ sub get_latlong_from_request {
         }
         my $x = $min_x + ( ( $max_x - $min_x ) / 2 );
         my $y = $min_y + ( ( $max_y - $min_y ) / 2 );
-        $latlong = [ $x, $y ];
+        $latlong = [ $y, $x ];
     }
 
     return $latlong;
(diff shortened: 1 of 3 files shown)


perllib/Open311/Endpoint/Integration/AlloyV2.pm
```diff
@@ -71,6 +71,11 @@ has config => (
     default => sub { $_[0]->alloy->config }
 );
 
+has group_in_service_code => (
```
dracos, line 74: If I've understood this right, should this not also have a change in post_service_request to not split the service_code on `_` into group/category? If it's more complicated, I think this needs a comment/documenting :)

perllib/Open311/Endpoint/Integration/AlloyV2.pm
```diff
@@ -683,11 +683,20 @@ sub get_latlong_from_request {
...
-        $latlong = [ $points[$half]->[0], $points[$half]->[1] ];
+        $latlong = [ $points[$half]->[1], $points[$half]->[0] ];
+    } elsif ( $geometry->{type} eq 'MultiLineString') {
+        my @points;
+        my @segments = @{ $geometry->{coordinates} };
+        for my $segment (@segments) {
+            push @points, @$segment;
+        }
+        my $half = int( @points / 2 );
+        $latlong = [ $points[$half]->[1], $points[$half]->[0] ];
```
dracos, line 706: How come the Polygon $latlong line doesn't need updating as well?

davea: Who says it doesn't? 🤦 Thanks!

perllib/Open311/Endpoint/Integration/AlloyV2.pm
davea: In actual fact `$group`/`$category` aren't used at all! So I'll get rid.

PR #125, opened 2020-07-08
Bartec fetch reports

struan:


perllib/Open311/Endpoint/Integration/Bartec.pm
```diff
@@ -270,6 +271,53 @@ sub get_service_request_updates {
...
+
+    my @requests;
+
+    my $updates = $self->_coerce_to_array( $response, 'ServiceRequest_Updates' );
+    for my $update ( @$updates ) {
+        my $res = $self->get_integration->ServiceRequests_Get( $update->{ServiceCode} );
+
+        next unless $res;
+        my $sr = $res->{ServiceRequest};
+        next if $sr->{ExternalReference} or not $sr->{ServiceStatus}->{Status} eq 'OPEN';
```
dracos, line 289: A comment here, perhaps? I assume ignoring ExternalReference as that means it's come from FMS, but what's the reason to ignore closed reports?

perllib/Open311/Endpoint/Integration/Bartec.pm
```diff
@@ -270,6 +271,53 @@ sub get_service_request_updates {
...
+
+        my $service_name = $sr->{ServiceType}->{Description};
+        next unless grep { uc $service_name eq uc $_ } keys %{ $self->allowed_services };
```
dracos, line 292: `next unless $self->allowed_services->{uc $service_name};` perhaps?

perllib/Open311/Endpoint/Integration/Bartec.pm
```diff
@@ -270,6 +271,53 @@ sub get_service_request_updates {
...
+        $service_name =~ s/(.)(.*)/\U$1\L$2/;
```
dracos, line 293: u and l work in strings, so no need for a regex: `$service_name = "\u\L$service_name";`

perllib/Open311/Endpoint/Integration/Bartec.pm
```diff
@@ -270,6 +271,53 @@ sub get_service_request_updates {
...
+
+        my $location = $res->{SOM}->dataof('//ServiceLocation/Metric');
+
+        my $date = $w3c->parse_datetime( $sr->{DateRequested} )->truncate( to => 'second' );
+        $date->set_time_zone('Europe/London');
+        my %args = (
+            service_request_id => $sr->{ServiceCode},
+            requested_datetime => $date,
+            updated_datetime => $date,
```
dracos, line 302: Does the history of the request need looking up here?

struan: Why do you ask that?

dracos: I wondered if you'd want to have the lastupdate time set at the FMS end, rather than going on the created date, but I don't think it would make any difference, would it, certainly not if you have to do an extra fetch per report.

PR #126, opened 2020-07-08
Bartec: add description as a note

struan:
Plus some other small tidy ups.


perllib/Integrations/Bartec.pm
```diff
@@ -373,7 +373,7 @@ sub ServiceRequests_Create {
         ServiceStatusID => $status_id,
         DateRequested => $time,
         ServiceTypeID => $values->{service_code},
-        serviceLocationDescription => $values->{description},
+        serviceLocationDescription => $values->{attributes}->{title} . "\n\n" . $values->{attributes}->{description},
```
dracos: I may well have missed something, but why is the front end concatenating the two fields on to the description and then this using it - why is this code here not just using the two attributes sent through to construct whatever is needed here?

perllib/Integrations/Bartec.pm
```diff
@@ -445,6 +493,11 @@ sub ServiceRequests_Detail_Get {
     return $self->_wrapper('ServiceRequests_Detail_Get', 0, @_);
 }
 
+sub ServiceRequests_Notes_Types_Get {
```
dracos, line 496: This function is never used, I think?

perllib/Open311/Endpoint/Integration/Bartec.pm
```diff
@@ -88,10 +91,16 @@ sub post_service_request {
     my $res = $integ->ServiceRequests_Create($service, $req);
     die "failed to send" unless $res->{ServiceCode};
 
+    my $sr = $integ->ServiceRequests_Get( $res->{ServiceCode} );
 
+    my $type = $integ->note_types->{ $integ->config->{note_types}->{report} };
+    $integ->ServiceRequests_Notes_Create({
+        srid => $sr->{ServiceRequest}->{id},
```
dracos, line 98: Is this the same as $res->{ServiceCode}? (just for my understanding)

perllib/Open311/Endpoint/Integration/Bartec.pm
```diff
@@ -88,10 +91,16 @@ sub post_service_request {
...
+        note_type => $type,
+        note => $args->{attributes}->{title} . "\n\n" . $args->{attributes}->{description},
```
dracos, line 100: I may well have missed something, but why is the front end concatenating the two fields on to the description and then this using it - why is this code here not just using the two attributes sent through to construct whatever is needed here?

perllib/Open311/Endpoint/Integration/Bartec.pm
struan: Sadly no, requests in Bartec have a service code and an id for for requests and which one you use depends on the API call.

perllib/Integrations/Bartec.pm
struan: It's used by `note_types` which is then used by `post_service_request`.

dracos: note_types calls _wrapper directly, it does not call this function.

struan: oh yes, so it does.

struan:
The new test for asset details wasn't expecting ReporterType was the test issue.

PR #128, opened 2020-07-15
[bartec] refactor skip request in request fetching

struan:
Move statuses to config and whole thing to a function.


davea:
(👍 assuming you can make Travis happy. Tests are fine locally.)

PR #131, opened 2020-07-28
Bartec bugfixes

struan:
Various small bugfixes


perllib/Open311/Endpoint/Integration/Bartec.pm
```diff
@@ -360,7 +360,23 @@ sub get_service_request_updates {
...
+
+    my $conf = $self->get_integration->config;
+
+    my $status = $conf->{status_map}->{ $update->{ServiceStatusName} };
+
+    if ($update->{ClosingCode} && $conf->{closing_code_map} &&
+        $conf->{closing_code_map}->{ $update->{ServiceStatusName} } &&
+        $conf->{closing_code_map}->{ $update->{ServiceStatusName} }->{ $update->{ClosingCode} } ) {
+        $status = $conf->{closing_code_map}->{ $update->{ServiceStatusName} }->{ $update->{ClosingCode} };
+    }
```
davea, line 379: I don't know if we need to be this careful here, something like this would appear to work as well:

```suggestion
    my $mapped = $conf->{closing_code_map}->{ $update->{ServiceStatusName} }->{ $update->{ClosingCode} };
    return $mapped if $mapped && $update->{ClosingCode};
```

PR #133, opened 2020-07-31
Bartec usrn lookup bug

struan:


perllib/Open311/Endpoint/Integration/Bartec.pm
```diff
@@ -231,12 +231,12 @@ sub get_uprn_for_street {
         usrn => $usrn,
     });
 
-    if (ref $premises->{Premises} eq 'ARRAY') {
-        for my $p ( @{ $premises->{Premises} } ) {
-            next unless $p->{Address}->{Address1} eq 'STREET RECORD';
+    $premises = $self->_coerce_to_array($premises, 'Premises');
```
dracos, line 234: It's fine, and I see is consistent, but I'd suggest the coerce_to_array calls (here and elsewhere) should probably be in the Integration code, so that its API is returning consistent output?

PR #135, opened 2020-08-05
[AlloyV2] change how we look up asset parent attribute

struan:
Rather than use the name of the attribute use the code of the thing we
are linking too which is in the options now. This should be a bit more
robust.


perllib/Integrations/AlloyV2.pm
```diff
@@ -130,7 +130,7 @@ sub get_parent_attributes {
 
     for my $att ( @{ $design->{design}->{attributes} } ) {
         $self->logger->debug($att->{name});
-        if ( $att->{name} eq $self->config->{parent_attribute_name} ) {
+        if ( $att->{options}->{code} && $att->{options}->{code} eq $self->config->{parent_attribute_name} ) {
```
dracos, line 133: Could be ($att->{options}->{code} || '') instead to avoid the warning but still be covered.

PR #144, opened 2020-08-21
[Northamptonshire] move extra status config to file

struan:
Rather than have code to set the reason for closure and status for non
standard statuses, do this in the config file for ease of expansion.

Part of mysociety/fixmystreet-commercial#1952


davea (approved):

struan:
merged in https://github.com/mysociety/open311-adapter/commit/7fe9b83001be1485de58c9880729973df830447f

PR #145, opened 2020-08-26
[AlloyV2] Use new attribute instead of item for geometry

davea:
See https://announcements.alloyapp.io/removal-of-item-title-subtitle-and-geometry-properties-(phase-2)-3Jrpcs


struan (approved):
👍

PR #148, opened 2020-09-02
[Alloy] sort all updates by date

struan:
make sure that if there are updates from defects and inspections we put
them in date order. Also slight change to test data to mimic Alloy
returning item logs in reverse date order.


chrismytton (approved):
📅

PR #149, opened 2020-09-03
[Alloy] utility script to get inspection status history

struan:
Script that takes an alloy itemId for an inspection or defect and prints
out the history of status changes. If you pass in an inspection id it
will also get the history of the related defect.


chrismytton (approved):
👍

Sorry for the slow review!

PR #153, opened 2020-09-22
[Central Beds] Add sending of updates

davea:
A simple change as Central Beds Symology doesn't require a different code for each status.


dracos (changes requested):
Would be good to have some sort of test of the Central Beds output.

davea:
Have added some tests for Central Beds in 9306505fc33fbb1b66da534383062fa96926b067.

dracos (approved):
🧪

PR #155, opened 2020-10-14
[Hackney] Fetching of completion photos

struan:
This will fetch any photos that were added as part of an update with a
status of in_progress or fixed.


t/open311/endpoint/json/alloyv2/inspect_search_photo.json
```diff
@@ -0,0 +1,85 @@
...
+  "page": 1,
+  "pageSize": 20,
+  "results": [
+    {
+      "itemId": "4027027",
+      "designCode": "designs_enquiryInspection_FixMyStreet",
+      "collection": "Live",
+      "start": "2020-10-11T00:00:05.123Z",
+      "end": "9999-12-31T23:59:59.999Z",
+      "geometry":{
```
davea, line 12: We should use the `attributes_itemsGeometry` attribute instead of the `geometry` property from now on, to mirror the Alloy API.

davea (changes requested):
Sorry to be pernickety, and I know it doesn't affect the code changes in this PR, but the `geometry` property has been removed from Alloy so we shouldn't include it in our fixtures anymore.

davea (approved):
📸  🚦  ✅

LouiseMySociety:
Agreed to close and will reopen if customer asked for it again.

PR #156, opened 2020-10-19
[Bexley] Handle private comments field

chrismytton:
This is to work with the changes at the FixMyStreet end to send private comments over when they are entered on the report form.

I tried putting the `sub _build_attributes` bit into the `BexleyPrivateComments` role, but you can't use the `SUPER` pseudo-class from a role because it refers to the package rather than the object ([source](https://perldoc.perl.org/perlobj#How-SUPER-is-Resolved)).

The FixMyStreet bits are in https://github.com/mysociety/fixmystreet/pull/3222

Part of https://github.com/mysociety/fixmystreet-commercial/issues/1850

### DIFF
--- a/perllib/Open311/Endpoint/Integration/UK/Bexley.pm
+++ b/perllib/Open311/Endpoint/Integration/UK/Bexley.pm
@@ -17,4 +17,20 @@ has integration_without_prefix => (
     default => 'Symology',
 );
 
+around post_service_request => sub {
+    my ($orig, $class, $integration, $report) = @_;
+
+    # Add private comments
+    my $private_comments = delete $report->{attributes}{private_comments};
+    $report->{description} .= "\nPrivate comments: " . $private_comments if $private_comments;
+
+    # Confirm overwrites the description with the one from the attributes,
+    # so add private comments there as well.
+    if ($report->{attributes}->{description}) {
+        $report->{attributes}->{description} .= "\nPrivate comments: " . $private_comments if $private_comments;
+    }
+
+    return $class->$orig($integration, $report);
+};
+
 __PACKAGE__->run_if_script;
--- a/perllib/Open311/Endpoint/Integration/UK/Bexley/ConfirmGrounds.pm
+++ b/perllib/Open311/Endpoint/Integration/UK/Bexley/ConfirmGrounds.pm
@@ -3,6 +3,13 @@ package Open311::Endpoint::Integration::UK::Bexley::ConfirmGrounds;
 use Moo;
 extends 'Open311::Endpoint::Integration::Confirm';
 
+use Open311::Endpoint::Service::UKCouncil::BexleyConfirm;
+
+has service_class  => (
+    is => 'ro',
+    default => 'Open311::Endpoint::Service::UKCouncil::BexleyConfirm'
+);
+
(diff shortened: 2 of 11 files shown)


davea (approved):
🔒 🔍 🗒️

perllib/Open311/Endpoint/Integration/UK/Bexley.pm
```diff
@@ -17,4 +17,14 @@ has integration_without_prefix => (
     default => 'Symology',
 );
 
+around post_service_request => sub {
+    my ($orig, $class, $integration, $report) = @_;
+
+    # Add private comments
+    my $private_comments = $report->{attributes}{private_comments};
```
dracos, line 24: This doesn't work once it tries to submit to Confirm as the attribute is still present and Confirm rejects it as "private_comments is not a valid attribute type code".
I think this would remove the attribute along with fetching it:
```suggestion
    my $private_comments = delete $report->{attributes}{private_comments};
```

dracos (changes requested):
Needs a fix from testing of Confirm submission (assume might apply to others as well, this should fix for all assuming this var isn't needed after this point)

perllib/Open311/Endpoint/Integration/UK/Bexley.pm
chrismytton: Thanks, have fixed in e7626f1.

dracos (approved):

chrismytton:
I've changed `hidden_field` to `server_set`, which I think is what it should have been in the first place.

With `hidden_field` it was creating an extra `private_comments` field in the FMS HTML, which was then causing problems down the line meaning the private comments weren't actually coming through.

chrismytton:
@dracos I've tried adding tests for the Bexley private comments bug with Confirm (wasn't sending private comments to Confirm because Confirm uses the description from the attributes). However, I can't figure out how to get this properly wrapped up in a test.

I've added one approach in b190d29bf932a0dbd6b820286afa05f92d4cd5f9, but it's failing for reasons I can't quite understand. I also tried using the same approach as `t/open311/endpoint/bexley.t` with using mocks of the `Open311::Endpoint::Integration::UK::Bexley::*` classes, but got mired in errors around missing attributes because the services in that test are generated with the simple `new_service` method, so don't have the `private_comments` attribute.

What am I missing? Is there something else I need to mock to get this approach to work?

dracos:
You're creating a Bexley endpoint, passing in a Bexley::ConfirmTrees config file (which a Multi subclass doesn't do anything with, so is ignored). So when it comes to look up the service in Bexley (first thing it does, as part of validating the received input), tries to pass it to its children, doesn't find anything, hence the error returned.
Looks like `t/open311/endpoint/confirm_upload.t` creates dummy Multi and a Confirm child in order to override this, as it says there bit fiddly, but can't think of an easier way off-hand, you could create a dummy child of Bexley and do the same. Or come up with a way to override a Bexley subclass config from within.

chrismytton:
@dracos Thanks! The `t/open311/endpoint/confirm_upload.t` example was enough for me to build on and get it all working 🎉

dracos (approved):
🔒

PR #157, opened 2020-10-19
[AlloyV2] Improve contact searching & creation; improve handling of large numbers of Alloy search results

davea:
Despite the query potentially having more than 2000 results, Alloy will raise an error if you try and fetch pages numbered beyond 100. In this case we recalculate the page size to stay below the 100 pages limit.

Also updates the contact search code to perform a narrower search on specific attributes.


perllib/Integrations/AlloyV2.pm
```diff
@@ -242,7 +243,9 @@ sub search {
     my $result_count = $stats->{result};
     return [] unless $result_count;
 
-    my $pages = int( $result_count / 20 ) + 1;
+    my $pageSize = 20;
+    my $maxPages = 100;
+    my $pages = min($maxPages, int( $result_count / $pageSize ) + 1);
```
struan, line 251: If the number of pages is > 100 we could divide the result_count by 100 and use that for the page size?

struan (approved):
I noticed this and was sure I'd fixed it but apparently not.

perllib/Integrations/AlloyV2.pm
davea: That's a better idea, yes! Done in 693b1d522849c9b623a1e0ecc59486ca1f29168c.

struan (approved):
👍

davea:
Merged in ef5fac66271d6706eb8d9a2bc128f41c02a50bfd.

PR #158, opened 2020-10-28
[Alloy] fetch completion photos from defects as well

struan:
NCC put the photos on the defect and not the inspection


davea (approved):
:shipit:

PR #159, opened 2020-10-29
[Bartec] allow extra questions for categories

struan:
Some categories in Bartec have extra data associated with them that is
send over as an extendedData field. Enable this to be configured and
then included in the service metadata, and then if present sent with an
enquiry.

Fixes mysociety/fixmystreet-commercial#2008


davea (approved):
👍

davea (approved):
👍

davea (approved):
🟢

PR #165, opened 2020-12-03
[Central Beds] Ensure SFTP SSH keys are accepted when fetching CSVs

davea:
Removes a gotcha when deploying on a new server for the first time.

### DIFF
--- a/bin/centralbeds/fetch_symology_csv
+++ b/bin/centralbeds/fetch_symology_csv
@@ -7,6 +7,10 @@ read_conf "$(dirname "$0")/../../conf/council-centralbedfordshire_symology.yml"
 
 FLAG=${1-}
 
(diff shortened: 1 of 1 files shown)


sagepe (commented):
I guess this will work, at least until they change the target server. `ssh-keyscan` will just add a "hostname cipher key" string, so you might find that in the future this fails due to mismatched keys (I see at present the hostname is a CNAME, so perhaps this could change in future).

Despite the fact that doing this automatically means we add a key without checking it, it does mean that we'd get alerted should the key change so I think this is marginally better security that just running `curl --insecure ...` instead.

From a strict security perspective neither of the above options are ideal - another option might be to verify the key and add it to the global `ssh_known_hosts` files on the application tier servers, that way it will be present wherever this service gets deployed.

PR #170, opened 2021-02-03
[Bartec] Add per-service Open311 keywords

davea:
- Makes it possible to add keywords to Open311 service definitions in the YAML config
 - Puts Bartec endpoint URLs in config for easier local debugging/testing


chrismytton (approved):
👍

PR #173, opened 2021-03-12
[Rutland] Add support for getting category groups from Salesforce

chrismytton:
This uses Rutland's new FixMyStreetInfoV2 endpoint which now has extra fields for specifying parent-child relationships.

The `parent` field indicates the parent category, AKA the group of the service.

There is also a `hasChildren` field which when set to "true" (string) indicates that the entry is a group and shouldn't be queried for attributes.

Part of https://github.com/mysociety/fixmystreet-commercial/issues/2206


dracos (approved):
🧛

PR #176, opened 2021-03-31
Memcached improvements

### LINKED ISSUE #113: Use different memcache keys for staging server
Several of the integrations use memcache to store frequently fetched data (usually the service list) and the key used is the same on staging and live. This means if we ever federated our memcache servers then there would be clashes.

Live and Staging should generate different key names.

dracos:
Mostly for the middle commit which had me very confused yesterday as to why one test, run by itself, was seeing a completely different test's service whitelist.
Fixes #113 (I think the namespace would have to be given once per config file it applies to, rather than centrally, as Integrations::* all have their own config files attached, but that probably doesn't matter?)

### DIFF
--- a/perllib/Integrations/Alloy.pm
+++ b/perllib/Integrations/Alloy.pm
@@ -2,7 +2,6 @@ package Integrations::Alloy;
 
 use DateTime::Format::W3CDTF;
 use Moo;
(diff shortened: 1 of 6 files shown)


struan (approved):
🧠

PR #177, opened 2021-03-31
[Echo] Take first value found rather than last

dracos:
For https://github.com/mysociety/fixmystreet/pull/3348
Also includes the parent change from another branch so it's hopefully easier to merge that in later.

### DIFF
--- a/perllib/Open311/Endpoint/Integration/Echo.pm
+++ b/perllib/Open311/Endpoint/Integration/Echo.pm
@@ -91,14 +91,13 @@ sub check_for_data_value {
 
 sub _get_data_value {
     my ($self, $name, $args, $request) = @_;
-    my $value;
     (my $name_with_underscores = $name) =~ s/ /_/g;
-    $value = $args->{$self->data_key_open311_map->{$name}} if $self->data_key_open311_map->{$name};
-    $value = $args->{attributes}{$name_with_underscores} if defined $args->{attributes}{$name_with_underscores};
-    $value = $self->default_data_all->{$name} if $self->default_data_all->{$name};
-    $value = $self->default_data_event_type->{$request->{event_type}}{$name}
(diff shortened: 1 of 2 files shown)


struan (approved):
👍

perllib/Open311/Endpoint/Integration/Echo.pm
```diff
@@ -115,15 +115,37 @@ sub post_service_request {
         my $row = { id => $type->{Id} };
         $row->{value} = $self->check_for_data_value($type->{Name}, $args, $request);
 
+        my %extra;
+        my $extra = 0;
```
struan, line 119: might be clearer if this was called `$extra_count`

t/open311/endpoint/echo.t
```diff
@@ -80,6 +87,20 @@ $soap_lite->mock(call => sub {
...
+                @children = ${$notes[0]->value}->value->value;
+                @child = ${$children[0]->value}->value;
+                is $child[0]->value, 1005;
+                is $child[1]->value, 2;
+                @child = ${$children[1]->value}->value;
+                is $child[0]->value, 1006;
+                is $child[1]->value, 1;
+                @child = ${$children[2]->value}->value;
+                is $child[0]->value, 1007;
+                is $child[1]->value, 12;
```
struan, line 102: should this not also check there's a second event here with a reason of 7?

struan (changes requested):
one variable name tweak and a bit more testing.

t/open311/endpoint/echo.t
dracos: It did that above when it runs the same test as the non-multi request, then this is extra to test the second one on top.

struan: ah yes, so it does.

Issue #29, opened 2019-01-02
Error in mapping Confirm code is only logged in main Apache error log


dracos:
The error line if there's no matching reverse mapping:
https://github.com/mysociety/open311-adapter/blob/d1583f120b146034077ad426e66ddb0daa2a671c/perllib/Open311/Endpoint/Integration/Confirm.pm#L432
is only being output to the main Apache error log. It defaults to 'open' if that's the case, so that's okay, but looks to have been 2,700 times in past 4 weeks, so we should probably either stop outputting it if we don't care about it happening, or have it log it somewhere better if we do. (Or perhaps this is wider issue of it not being in app-specific error log.)

Issue #20, opened 2018-05-31
Enable logging of requests send to external systems


struan:
To help with debugging failed reports it would be helpful to be able to see the requests made to external systems so there should be some way to do this.

PR #178, opened 2021-04-01
Logging improvements

dracos:
Does some tidying of config set up and the like, but the main part is the bit that makes sure any error during the actual API code is logged as well as sent to the client (so making universal those integrations that had a `log_and_die` function), and making sure everything uses the logger rather than STDERR or warn or whatever.

This should stop anything going to the main Apache log, everything should go to the log file set in the global general.yml at the level set there.

Fixes https://github.com/mysociety/open311-adapter/issues/29
Might fix https://github.com/mysociety/open311-adapter/issues/20 (it will log all SOAP transport with debug logging, and otherwise log all SOAP errors). Other logging can be done by the integration.


davea (approved):
👍

PR #179, opened 2021-04-07
[Oxfordshire] Alloy integration

davea:
A few smallish changes to support report/update sending/fetching for Oxfordshire.


perllib/Open311/Endpoint/Integration/AlloyV2.pm
```diff
@@ -858,7 +858,7 @@ sub get_defect_category {
         my $type;
 
         for my $att (@attributes) {
-            if ($att->{attributeCode} =~ /DefectType/ ) {
+            if ($att->{attributeCode} =~ /DefectType/ || $att->{attributeCode} =~ /DefectFaultType/) {
```
struan, line 861: I wonder if it makes sense to move this into the configuration?

t/open311/endpoint/uk.t
```diff
@@ -76,4 +75,21 @@ foreach ($endpoint->plugins) {
...
+
+foreach ($endpoint->plugins) {
+    my $integ = $_->get_integration;
+    my $name = delete $config_filenames{ref($_)};
+    is $integ->config_filename, $name;
+    is basename($integ->config_file), "council-$name.yml";
+}
+
+is_deeply \%config_filenames, {};
+
```
struan, line 95: I wonder if it might make sense to refactor these essentially identical tests

struan (approved):
Presumably the first commit is no longer WIP ;)

Some other small comments but not urgent.

davea (commented):
Your changes look good @dracos 

> Is the parent class function wrong, then?

Hackney doesn't appear to override `_get_defect_inspection_parents`, but I can't think of why it shouldn't. So perhaps the parent function is wrong.

dracos:
@davea Could you just check over 7d5483f61e182cc66a217b34c27eab9d77bd6d3b which is me putting that defect parent code in the main Alloy code assuming it should apply to all. No other changes since your approval.

PR #181, opened 2021-04-12
[Central Beds] Send report title to enquiry Location field in Symology

davea:
For https://github.com/mysociety/fixmystreet-commercial/issues/2231


perllib/Open311/Endpoint/Integration/Symology.pm
```diff
@@ -129,7 +129,7 @@ sub process_service_request_args {
 
     # We need to bump some values up from the attributes hashref to
     # the $args passed
-    foreach (qw/fixmystreet_id easting northing UnitID RegionSite NSGRef contributed_by/) {
+    foreach (qw/fixmystreet_id easting northing UnitID RegionSite NSGRef contributed_by report_url description/) {
```
dracos, line 132: I don't think this line change is needed - neither of the new variables are used/needed in the $request objject (a "Description" value is used, so the lowercase version might be confusing).

The $args->{report_url} check a few lines down could presumably start using $args->{attributes}->{report_url} instead now that it could be present.

dracos (changes requested):
One change not needed, I think, is all

perllib/Open311/Endpoint/Integration/Symology.pm
davea: You're right that they're not used in $request, but left alone they get appended to the `Description` by the loop on L163. Maybe the better thing to do would be to remove them in `CentralBedfordshire::process_service_request_args`, given they're specific to that integration.

dracos: Oh yes! Yes, I guess so, easy to move back if it becomes more generic at some point.

dracos (approved):

PR #182, opened 2021-04-15
[Bexley] use Symology Event Type as external status

struan:
Use the Event Type column as the external status code if present in the
update CSV.

Fixes mysociety/fixmystreet-commercial#2279

### DIFF
--- a/perllib/Open311/Endpoint/Integration/UK/Bexley/Symology.pm
+++ b/perllib/Open311/Endpoint/Integration/UK/Bexley/Symology.pm
@@ -55,13 +55,13 @@ sub _row_status {
             'fixed'
         } elsif ($action_due =~ /^[NS][1-6]$/) {
             'in_progress'
-        } elsif ($action_due eq 'IR') {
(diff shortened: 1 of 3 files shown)


perllib/Open311/Endpoint/Integration/UK/Bexley/Symology.pm
```diff
@@ -71,6 +71,12 @@ sub _row_status {
     };
 }
 
+sub _row_external_status_code {
+    my ($self, $row, $status) = @_;
+    return undef unless $status eq 'not_councils_responsibility';
```
davea, line 76: `$status` may be `undef` here which causes an uninitialized value warning, how about:

```suggestion
    return undef unless $status && $status eq 'not_councils_responsibility';
```

davea (changes requested):
One warning to squash is all.

davea (approved):
LGTM!

struan:
Extra mapping of categories added

chrismytton (approved):
🟢

PR #183, opened 2021-04-20
Switch O::E::I:: config to endpoint_config.

dracos:
This partially reverts commit c55f7eed9. Web::Simple::Application has an attribute called `config`, so we cannot use that in a subclass, as it has already been set. Even if it didn't, I'm not sure anything was actually setting config_filename in these classes to be used.

The test move to use a yml file rather than mock config directly would be enough to catch the first of these, because config was then not set at all when running the test.

### DIFF
--- a/perllib/Role/EndpointConfig.pm
+++ b/perllib/Role/EndpointConfig.pm
@@ -1,0 +1,34 @@
+# Very similar to Role::Config, but to be used by subclasses of Web::Simple
+# (such as any Open311::Endpoint::Integration::...) as that sets a config
+# attribute
+
+package Role::EndpointConfig;
+
+use Path::Tiny;
+use YAML::XS qw(LoadFile);
+use Moo::Role;
+
+has config_file => (
+    is => 'lazy',
+    coerce => sub { path($_[0]) },
+);
+
+sub _build_config_file {
+    my $self = shift;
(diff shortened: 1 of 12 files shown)


perllib/Role/EndpointConfig.pm
```diff
@@ -0,0 +1,34 @@
...
+}
+
+has endpoint_config => (
+    is => 'lazy',
+    isa => sub { die "not a hashref" unless ref $_[0] eq 'HASH' },
+);
+
+sub _build_endpoint_config {
+    my $self = shift;
+    return {} if !$self->config_file->is_file && $ENV{TEST_MODE};
```
struan, line 29: Is there a reason we're not turning on the YAML::XS::Boolean thing from Role::Config here?

struan (approved):
🐮

perllib/Role/EndpointConfig.pm
dracos: I was spooked by your comment `we could possibly use local $YAML::XS::Boolean = "JSON::PP" in the Config module to get round all this but not sure if that would break something else.` , and yet the Config module has had that change, I see :) And the code I was refactoring together didn't have it. I think it only matters if you need to pass the parsed JSON out again (which was needed for Alloy), Symology checks

struan: that sounds like a reason :)

PR #186, opened 2021-04-23
[Bexley] script to process update CSV to FMS values

struan:
Take a symology update CSV and produce a CSV with the updates values
required to import it into FixMyStreet.


dracos (approved):
Looks nice :) Could it perhaps output a JSON array, would that be slightly easier to read at the FMS end? Probably not much in it, but I did briefly wonder about UTF-8 characters, etc

dracos (approved):

PR #187, opened 2021-05-04
[Confirm] Add EnquiryClassCode support.

dracos:


davea (approved):
Might be nice to have a little test too?

PR #188, opened 2021-05-13
[Rutland] Pass through Salesforce HTML hint text in attributes

chrismytton:
Rutland's Salesforce now includes an `html` field on categories and groups which Rutland want to display on the FMS front end. This change puts the HTML from Salesforce for both the category and the category's group into the service's attributes so that FMS can pull it out at the other end.

The PR to output this text on FixMyStreet is here: https://github.com/mysociety/fixmystreet/pull/3434.

Part of https://github.com/mysociety/societyworks/issues/2206

## Notes to reviewer

I originally tried doing this with new top-level elements, but that didn't quite feel right and broke lots of tests. So I went with putting it in the attributes in the end, even though it doesn't feel _quite_ right. You can see the code I wrote for that here: https://github.com/mysociety/open311-adapter/compare/rutland-salesforce-groups-support...rutland-salesforce-html-hint-text.

## Notes to merger

- [ ] Merge https://github.com/mysociety/open311-adapter/pull/173 before this PR, as this PR builds on top of that one

### DIFF
--- a/perllib/Open311/Endpoint/Integration/SalesForce.pm
+++ b/perllib/Open311/Endpoint/Integration/SalesForce.pm
@@ -207,6 +207,22 @@ sub service {
     my ($self, $id, $args) = @_;
 
     my $meta = $self->get_integration->get_service($id, $args);
+    my @services = $self->get_integration->get_services($args);
+
+    my %service_lookup = map { $_->{serviceid} => $_ } @services;
+    my $srv = $service_lookup{$id};
+    my $parent;
+    my $hint = '';
+    my $group_hint = '';
+
+    if ($srv) {
+        $parent = $service_lookup{$srv->{parent}};
+        $hint = $srv->{html};
+    }
+
+    if ($parent) {
+        $group_hint = $parent->{html};
+    }
 
     my $service = Open311::Endpoint::Service::UKCouncil::Rutland->new(
         service_name => $meta->{title},
@@ -232,6 +248,25 @@ sub service {
         my $attrib = Open311::Endpoint::Service::Attribute->new(%options);
         push @{ $service->attributes }, $attrib;
     }
+
+    my %options = (
+        required => 0,
+        variable => 0,
+        datatype => 'string',
+        automated => 'server_set',
+    );
+
+    push @{ $service->attributes }, Open311::Endpoint::Service::Attribute->new(
+        code => 'hint',
+        description => $hint,
+        %options,
+    );
+
+    push @{ $service->attributes }, Open311::Endpoint::Service::Attribute->new(
+        code => 'group_hint',
+        description => $group_hint,
+        %options,
+    );
 
     return $service;
 }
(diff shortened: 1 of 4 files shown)


perllib/Open311/Endpoint/Integration/SalesForce.pm
```diff
@@ -207,6 +207,22 @@ sub service {
     my ($self, $id, $args) = @_;
 
     my $meta = $self->get_integration->get_service($id, $args);
+    my @services = $self->get_integration->get_services($args);
```
dracos, line 210: Worth adding a memcache use to `get_services` (like a few other integrations do), even if only for a few minutes, to save this being requested once per service every time the populate loops through all the services.

perllib/Open311/Endpoint/Integration/SalesForce.pm
```diff
@@ -233,6 +249,24 @@ sub service {
         push @{ $service->attributes }, $attrib;
     }
 
+    my %options = (
+        required => 0,
+        datatype => 'string',
+        automated => 'server_set',
+    );
+
+    push @{ $service->attributes }, Open311::Endpoint::Service::Attribute->new(
```
dracos, line 258: Should you push these only if set? Guess doesn't make much difference.

perllib/Open311/Endpoint/Integration/SalesForce.pm
```diff
@@ -233,6 +249,24 @@ sub service {
         push @{ $service->attributes }, $attrib;
     }
 
+    my %options = (
```
dracos, line 252: Needs a variable => 0 too, I think.

dracos (changes requested):
Couple of small suggestions is all.

Regarding your comment - the reason it broke lots of tests was because hint/group_hint were made mandatory everywhere. If you'd made them optional in Schema.pm and changed Endpoint.pm to only include them if set (e.g. `$service->hint ? (hint => $service->hint) : (),`) then all the existing tests pass. But I agree, at least at this point doing it as attributes is slightly easier - can be dealt with by existing code hooks at the FMS end as you have done, for one thing - but could be changed in future if others want to do it, I guess. And if some other standard Open311 client is reading this data, it would see these notices anyway, rather than having to know about the extra fields.

perllib/Open311/Endpoint/Integration/SalesForce.pm
chrismytton: Good idea! Have tidied up the existing memcached code and added caching to the `get_services` call 👍

perllib/Open311/Endpoint/Integration/SalesForce.pm
chrismytton: Yeah I don't think it really matters either way. I guess it might be slightly clearer if the attributes are always there, to make it obvious that they _could_ be populated in the future.

dracos:
I'll review this on Monday, but just to note that Memcached was already factored out to a Role on master back in March, sorry!

dracos:
Your refactoring is identical to the one already there, at least :-)

dracos (approved):
(Once rebased on current code ;) )

chrismytton:
@dracos Haha, oh dear. I should have rebased onto `master`, that will teach me!

PR #190, opened 2021-06-01
[Central Beds] Send photo URLs to field in Symology

davea:
For https://mysocietysupport.freshdesk.com/a/tickets/1076


chrismytton (approved):
📸

dracos:
As this has been worked out/fixed for Camden, should we merge this now?

davea:
Yes, I think that would be a good idea.

PR #191, opened 2021-06-02
[Bartec] Improve report sending resilience

davea:
- Make sure report content is properly escaped
 - Stop report sending requests failing after Bartec ServiceRequest has been created, leading to duplicate reports being raised.

For https://github.com/mysociety/societyworks/issues/2391

### DIFF
--- a/perllib/Open311/Endpoint/Integration/Bartec.pm
+++ b/perllib/Open311/Endpoint/Integration/Bartec.pm
@@ -3,6 +3,7 @@ package Open311::Endpoint::Integration::Bartec;
 use JSON::MaybeXS;
 use Path::Tiny;
 use MIME::Base64;
+use Try::Tiny;
 
 use Integrations::Bartec;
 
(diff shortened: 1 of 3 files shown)


t/open311/endpoint/bartec.t
```diff
@@ -579,7 +579,7 @@ subtest "check send report with extended info" => sub {
         token => 'ABC=',
         ServiceRequestID => '0001',
         NoteTypeID => 11,
-        Note => "a title\n\na description",
+        Note => "a title\n\na description & some text",
```
struan, line 582: I'm a bit confused as to what the encode data is doing given it doesn't seem to do anything with the ampersand, or does that happen after this?

perllib/Open311/Endpoint/Integration/Bartec.pm
```diff
@@ -172,19 +173,41 @@ sub post_service_request {
...
+        $self->logger->warn("failed to attach note to ServiceRequest: " . $res->{ServiceCode} . " (FMS ID " . $args->{attributes}->{fixmystreet_id} . ")");
+    };
 
-    return $self->new_request(
-        service_request_id => $res->{ServiceCode}
-    );
+
+    try {
+        if ( @{ $args->{media_url} }) {
+            $self->upload_urls($sr->{ServiceRequest}->{id}, $args); # XXX not sure ServiceCode is correct
```
struan, line 202: do these XXX things still apply?

struan (approved):
Looks fine, couple of minor comments but seems reasonable.

t/open311/endpoint/bartec.t
davea: It's been decoded again by the time we're running this test, so this is testing that a) we reach this point at all and b) the decoded data is correct.

perllib/Open311/Endpoint/Integration/Bartec.pm
davea: Nope, will remove.

PR #192, opened 2021-06-02
[Echo] ignore emails on anonymous reports

struan:
don't include the user's email if contributed_as is set to
anonymous_user


davea (approved):
Looks good 👍

PR #196, opened 2021-09-22
[Alloy] Switch from start field to attribute.

dracos:
The start field is now being set to 9999-12-31.
Fixes https://github.com/mysociety/societyworks/issues/2589


chrismytton (approved):
📆

PR #198, opened 2021-10-11
[Confirm] Remove file locking.

dracos:
This data is currently held over NFS and we were seeing stuck flock calls.
These files are not written to, only read or removed, and the script only
runs on one server anyway, it would still have a race condition if run on
more than one even with shared locking.


sagepe (approved):
Looks good, switch from `slurp_utf8` that uses `LOCK_EX` to `openr_utf8` that uses `LOCK_SH` - no need for exclusive lock as we're simply reading out the JSON.

PR #202, opened 2021-10-25
[Shropshire] Add Shropshire.pm file

MorayMySoc:
Shropshire.pm file for Shropshire integration


dracos:
This was merged as 770821b, looks like.

PR #209, opened 2021-12-14
[Confirm] Allow _extra on whitelist codes to work and map to same Confirm service/subject

dracos:
Fixes https://github.com/mysociety/societyworks/issues/2797 (once config is updated, see below).

Current live config has one issue without this, in that it has two `RM_FLO` but only one gets created.
Adapted config for changes currently inactive in live would be:

```
Trees:
  RM_TREE_1: Dead/diseased tree
  RM_TREE_2: Other tree issue
  RM_TREE_3: Overhanging branch/tree
  RM_TREE_4: Tree roots raising a footpath
  RM_TREE_5: Tree stumps
  SLRM_SLD_tree: Tree blocking a street light
  ITS_SMP_tree: Tree blocking traffic signal/sign
```

### DIFF
--- a/perllib/Open311/Endpoint/Integration/Confirm.pm
+++ b/perllib/Open311/Endpoint/Integration/Confirm.pm
@@ -30,13 +30,15 @@ Subclasses must override this or no Open311 services will be published!
 Returns a hashref which groups services together and optionally provides
 an overridden name for each service.
 
-For example, the following hashref will publish 5 Open311 services:
+For example, the following hashref will publish 7 Open311 services:
 
 {
     'Roads' => {
         RO_PH => 1,
         RO_GB => 1,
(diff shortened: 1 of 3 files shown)


dracos:
"This won't actually fix the populate-service cron though, as there will still be multiple services with the same code coming through"
- perhaps this isn't the right solution, and we should somehow map to different service codes for FMS and then map back when the reports come through. But that also feels confusing, so I'm hopeful the way I've done it here is the best way.

davea (approved):
This looks fine to me. We *could* rejig it so that it always outputs a unique name per code (or vice versa) as you suggest but that kind of feels like FMS's problem and what you've done here is cleaner.

davea (approved):

PR #210, opened 2021-12-16
[Salesforce] Factor common functions together.

dracos:
This means both options can use credentials or private keys.
For https://github.com/mysociety/societyworks/issues/1943


chrismytton (approved):
Woo, awesome, thanks for sorting this out!

PR #211, opened 2022-01-07
[Alloy] Some performance changes

dracos:
These are all actually present on the live site at present, this tidies them up a bit.

### DIFF
--- a/perllib/Integrations/AlloyV2.pm
+++ b/perllib/Integrations/AlloyV2.pm
@@ -33,6 +33,7 @@ sub api_call {
 
     my $ua = LWP::UserAgent->new(
         agent => "FixMyStreet/open311-adapter",
+        timeout => 5*60,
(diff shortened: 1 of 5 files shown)


davea (approved):
Code looks good, nice to improve things a little bit! I wonder about the effects of ignoring failing `reconstruct` calls though - are updates potentially going to get missed? Not failing the entire Open311 request is of course preferable :) Would be good to get to the bottom of what's going wrong.

dracos:
"are updates potentially going to get missed?" - theoretically, though looking at the examples on which this happens it's the creation log event and perhaps the one after, later are fine. So I don't think it should have any real effect on anything, in comparison to the opposite. Alloy are investigating the issue. I guess therefore might want to be alerted if it does happen again, I guess should add some logging of something, though would we spot it...

PR #214, opened 2022-03-04
[Confirm] Include completion photo in update.

dracos:
Adds a new ‘completion_statuses’ config key which controls when Confirm should be queried for a job completion photo when fetching Open311 updates.

Changes from when you wrote it: Removed memcached code, removed login token code (oauth works), tidied up API call functions and moved bits round, made Lincolnshire-specific function, added tests.

Have tested locally with both production and development, and appears to work :)

### DIFF
--- a/perllib/Integrations/Confirm.pm
+++ b/perllib/Integrations/Confirm.pm
@@ -1,5 +1,7 @@
 package Integrations::Confirm;
 
+use v5.14;
+use warnings;
 use SOAP::Lite;
 use Exporter;
 use DateTime::Format::W3CDTF;
@@ -70,6 +72,16 @@ has 'customer_type_code' => (
 has 'server_timezone' => (
     is => 'lazy',
     default => sub { $_[0]->config->{server_timezone} }
+);
+
+has completion_statuses => (
+    is => 'lazy',
+    default => sub { $_[0]->config->{completion_statuses} || [] }
+);
+
+has base_url => (
+    is => 'lazy',
+    default => sub { $_[0]->config->{base_url}  }
 );
 
 has oauth_token => (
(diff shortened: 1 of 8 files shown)


perllib/Integrations/Confirm.pm
```diff
@@ -507,20 +519,50 @@ sub upload_enquiry_documents {
...
+sub documents_for_job {
+    my ($self, $job_id) = @_;
+    my $data = $self->json_web_api_call("/jobs/$job_id");
+    return $data->{documents} || [];
+}
+
+sub get_job_photo {
+    my ($self, $job_id, $photo_id) = @_;
+    my $response = $self->web_api_call("/documents/0/JOB/$job_id/$photo_id") or return;
+    return ( $response->header('Content-Type'), $response->decoded_content );
```
davea, line 565: For safety/paranoia maybe we should check the `Content-Type` is a JPEG (and not e.g. PDF) before returning content.

perllib/Open311/Endpoint/Integration/UK/Lincolnshire.pm
```diff
@@ -24,5 +24,9 @@ sub process_service_request_args {
     return $args;
 }
 
+sub photo_filter {
+    my ($self, $doc) = @_;
+    return $doc->{fileName} =~ /jpe?g/i && $doc->{documentNotes} =~ /after/i;
```
davea, line 29: `documentNotes` isn't always defined (e.g. job 3052603 doc 22) which generates a `Use of uninitialized value in pattern match (m//)` warning here.

davea (changes requested):
A couple of small things, but overall this works well! Glad it didn't need that extra token in the end.

One concern is the impact this will have on the speed of fetching updates - testing locally against a production environment and the response time for 2 hours of updates (the default) went from ~3s to over 5 minutes 🤯

PR #215, opened 2022-03-10
Allow any service attribute through by default

chrismytton:
Previously this was only set on the Echo integration (see https://github.com/mysociety/open311-adapter/commit/706644b47883c57e0373ca4e27559cc0088e0c4b), but this is probably a sensible default generally, in the spirit of being liberal with what you receive.

I'm printing out a warning to the logfile if an attribute is unrecognised, but these are suppressed if the service has the allow_any_attributes flag set.

This is a roundabout way of fixing the error that Bucks is getting on staging currently, as that's an issue with the new grass cutting question not being recognised.

The change to `.github/workflows/default.yml` was necessary because calling methods on `$self->logger` fails without a `conf/general.yml` present.

### DIFF
--- a/perllib/Integrations/Confirm.pm
+++ b/perllib/Integrations/Confirm.pm
@@ -292,7 +292,8 @@ sub NewEnquiry {
         datetime => 'EnqAttribDateValue',
     };
 
-    for my $code (keys %{ $args->{attributes} }) {
+    for my $code (map { $_->code } @{ $service->attributes }) {
+        next unless exists $args->{attributes}->{$code};
         next if grep {$code eq $_} ('easting', 'northing', 'fixmystreet_id', 'closest_address');
         my $value = substr($args->{attributes}->{$code}, 0, 2000);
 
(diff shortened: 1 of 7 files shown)


dracos (changes requested):
One suggestion on how to not need the blank config file.

More generally, though, I'm not sure this will work as is - it looks like Integrations::Confirm loops through all attributes (line 295) and sends them all through to Confirm, which I assume (but haven't tested) will error when given data it is not expecting (going by the comment in that loop). I think this is easy to fix, though,in that you have `$service->attributes` available so you could loop through that rather than all attributes, I think?

Salesforce also does the same, don't know if it would handle unknown data but again could look at `$service` similarly. The rest look okay, I think but worth you checking.

chrismytton:
@dracos Have pushed a few fixup commits to address the Confirm and Salesforce issues you mentioned. I had a look through the other integrations but couldn't see anything obvious, though tricky to be certain!

dracos (approved):
Woo, looks good

PR #217, opened 2022-03-22
[Hackney] Alloy environmental services integration

davea:
Refactors the Hackney integration to be a `::Multi` with two AlloyV2 integrations.


perllib/Open311/Endpoint/Integration/AlloyV2.pm
```diff
@@ -393,15 +393,16 @@ sub post_service_request_update {
     my ($self, $args) = @_;
 
     my $resource_id = $args->{service_request_id};
-    my $inspection = $self->alloy->api_call(call => "item/$resource_id")->{item};
+    my $item = $self->alloy->api_call(call => "item/$resource_id")->{item};
 
-    my $attributes = $self->alloy->attributes_to_hash($inspection);
-    my $updates = $attributes->{$self->config->{inspection_attribute_mapping}->{updates}} || '';
+    my $attributes = $self->alloy->attributes_to_hash($item);
+    my $attribute_code = $self->config->{inspection_attribute_mapping}->{updates} || $self->config->{defect_attribute_mapping}->{updates};
```
dracos, line 399: This is fine - should it actually use defect_attribute_mapping *instead* of inspection_attribute_mapping, rather than as well? Bit confused about defect/inspection thing, different councils using them differently?

dracos (changes requested):
A few small requests/queries; would be good to test the new `defect_status` function (and indeed any basic smoke tests for hackney at all!).

perllib/Open311/Endpoint/Integration/UK/Hackney/Environment.pm
```diff
@@ -16,4 +16,29 @@ around BUILDARGS => sub {
...
+    my $reason = $defect->{$self->config->{defect_attribute_mapping}->{cancellation_reason}};
+    if ( $task_status && $reason ) {
+        $task_status = $task_status->[0] if ref $task_status eq 'ARRAY';
+        $reason = $reason->[0] if ref $reason eq 'ARRAY';
+
+        $return_status = $reason_mapping->{$reason} if $task_mapping->{$task_status} eq 'Cancelled';
+    }
+
+    unless ( $return_status ) {
+        $status = $status->[0] if ref $status eq 'ARRAY';
```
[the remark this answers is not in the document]
davea: Yeah good idea, makes it a little tidier.

perllib/Open311/Endpoint/Integration/AlloyV2.pm
davea: Yes, this is because some councils want defects to be updated and some want inspections to be updated instead. It's only ever one or the other so order doesn't matter (I think?)

dracos (changes requested):
Changes to requests all look fine, still bit worried that the new defect_status function isn't tested (or as I said, any of Hackney).

dracos (approved):
"increase coverage by 0.70%." 🎉  Woo!

Issue #204, opened 2021-11-08
Confirm photo upload failure shouldn't fail entire request [2hrs]


davea:
Any failure or error in actions that take place after the successful creation of the enquiry in Confirm (such as storing photo JSON for later upload) must not fail the entire request. If that happens a duplicate enquiry will be created each time FMS retries the Open311 POST Service Request.

dracos:
@davea Is the easiest solution here to move store_enquiry_documents above perform_request? So if the store fails and errors for some reason,the request won't be created. Opposite way round there isn't an issue, if Confirm fails and store_enquiry_documents is run more than once by retries.

jacquelinebylau:
Toggl: https://plan.toggl.com/#timeline/317959/tasks/33843681/embed

PR #219, opened 2022-04-29
Wrap Confirm document storage in an eval

nephila-nacrea:
Fixes https://github.com/mysociety/open311-adapter/issues/204.

This fix prevents the entire upload of a report from being marked as a failure if the initial report creation succeeds but the later document upload fails.


davea (approved):
Nicely done! 👍

PR #220, opened 2022-05-05
[Central Beds] Replace curl with sftp.

dracos:
curl is not working on the new server, returning a 79 (SSH) error.


davea (approved):
Looks good to me. Is `sshpass` installed by default on our infrastructure? If not do we need to make sure it's a dependency of `open311-adapter` vhosts?

PR #223, opened 2022-07-22
[Bexley] Add next action to updates

MorayMySoc:
https://github.com/mysociety/societyworks/issues/3059

* Adds check for nsg_ref parameter and, if it exists, look up relevant code and add it to the Next Action

### DIFF
--- a/perllib/Open311/Endpoint/Role/mySociety.pm
+++ b/perllib/Open311/Endpoint/Role/mySociety.pm
@@ -183,6 +183,13 @@ sub POST_Service_Request_Update_input_schema {
         }
     }
 
+    # Allow nsg_ref through for Bexley XXX
+    if (($args->{jurisdiction_id} || '') eq 'bexley') {
+        if ($args->{'nsg_ref'}) {
+	        $attributes->{optional}{'nsg_ref'} = '//str';
+        }
+    }
+
     return $attributes;
 }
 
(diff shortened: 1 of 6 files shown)


davea (commented):
Looking good so far

perllib/Open311/Endpoint/Role/mySociety.pm
```diff
@@ -173,6 +173,7 @@ sub POST_Service_Request_Update_input_schema {
             service_request_id_ext => '//num',
             public_anonymity_required => Open311::Endpoint::Schema->enum('//str', 'TRUE', 'FALSE'),
             service_code => $self->get_identifier_type('service_code'),
+            nsg_ref => '//str',
```
davea, line 176: Rather than adding a new top-level attribute I think it would be better/cleaner to follow the same pattern as for Oxfordshire just below on L180 and use `attribute[nsg_ref]`.

davea (changes requested):
One small change is all!

davea (approved):
👍

MorayMySoc:
Sorry - issue with the jurisdiction_id which my test cheated into working, pointed out by @dracos after tried on staging and didn't work.

davea (approved):
Oops, sorry for missing that. Looks good!

PR #226, opened 2022-09-06
[Bartec] Correctly handle required extra fields that are strings, not lists

davea:
Previously the code assumed that all Bartec extended data fields were lists. This was fine until we encountered a field that was actually a string and also mandatory. The schema validator threw an error due to the situation where there was field that requires a value but has no values to choose from.

This change ensures the datatype of the Open311 attribute is only set to singlevaluelist if there are values to be picked from.


dracos (approved):
Looks good :)

Issue #208, opened 2021-12-14
Alloy API key changes [15hrs]


davea:
Yotta are [changing the way API keys are managed in Alloy](https://announcements.alloyapp.io/new-api-key-management-xGuAg).

The important bit for us:

> Important: The way you pass the API key is changing!
>
> The new API keys will need to be passed according to the OAuth 2.0 Bearer token format (RFC 6750). That is a request header named "Authorization" will need to be included, which would look like this:
> 
> `Authorization: Bearer APIKEYVALUEHERE`
>
> This change means that the API keys provided by the old and new mechanisms are not interchangeable and old keys will not be accepted via the Authorization header and new keys will need to be generated.  

Deadline:

> Expected Release Date
> 13th January 2022
>
> Phase 2: Retirement of Existing Token Method
> December 2022

lizettal:
To be scheduled before December 2022

dracos:
`alloy-new-key` branch has the code change in; the key switch sounds like it'll be the annoying thing!

lizettal:
Scheduled for June 2022 https://plan.toggl.com/#timeline/317959/tasks/33070599/embed

dracos:
To get a new API key (here for Oxfordshire live):
```
CUSTOMER_ID=customers_oxfordshireLive_61af3b1c34277ed1048abce8
# Get a user token (doesn't do much)
curl -H"Content-Type: application/json" \
  --data '{"email":"community@mysociety.org","password":"..."}' \
  'https://api.uk.alloyapp.io/api/session'
# Get a customer token from user token
curl -H"Authorization: Bearer <key from step 1>" \
  -H"Content-Type: application/json" \
  --data '' \
  'https://api.uk.alloyapp.io/api/session/customer/$CUSTOMER_ID'
# Get an API key from customer token
curl -H"Authorization: Bearer <key from step 2>" \
  -H"Content-Type: application/json" \
  --data '{"label":"FixMyStreet","enabled":true}' \
  'https://api.uk.alloyapp.io/api/api-key'
```

dracos:
I have new API keys for Oxfordshire Live/Test, Hackney Highways Live/Test, Hackney Estates Live/Test and Hackney  TradeTest.

dracos:
This is live for tilma, will do this next week.

PR #227, opened 2022-09-13
[Alloy] Use new key method

dracos:
New keys have been fetched, so would need putting in server config as this was deployed.
Fixes https://github.com/mysociety/open311-adapter/issues/208

### DIFF
--- a/perllib/Integrations/AlloyV2.pm
+++ b/perllib/Integrations/AlloyV2.pm
@@ -39,10 +39,9 @@ sub api_call {
     $method = $body ? 'POST' : 'GET' unless $method;
     my $uri = URI->new( $self->config->{api_url} . $call );
 
-    $args{params}->{token} = $self->config->{api_key};
-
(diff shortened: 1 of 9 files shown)


perllib/Integrations/AlloyV2.pm
```diff
@@ -39,10 +39,9 @@ sub api_call {
...
     my $uri = URI->new( $self->config->{api_url} . $call );
 
-    $args{params}->{token} = $self->config->{api_key};
-
     $uri->query_form(%{ $args{params} });
     my $request = HTTP::Request->new($method, $uri);
+    $request->header(Authorization => 'Bearer ' . $self->config->{api_key});
     if ($args{is_file}) {
         $request = HTTP::Request::Common::POST(
             $uri,
```
davea, line 47: I think this request needs the header setting too.

davea (changes requested):
One more request instance that needs the new header.

PR #228, opened 2022-09-27
[Lincolnshire] Pass customer details through from Confirm

chrismytton:
Lincolnshire want reports created in Confirm to be associated with the correct user so that the user can receive updates from FixMyStreet.

This change get the customer name and email from Confirm and passes them to FMS in the GET Service Requests call.

FixMyStreet change that uses this data: https://github.com/mysociety/fixmystreet/pull/4110

Part of https://github.com/mysociety/societyworks/issues/3079

## Notes to deployer

- [x] Add `include_private_customer_details: 1` to the Lincs Confirm config before deploying this change

### DIFF
--- a/conf/council-lincolnshire_confirm.yml-example
+++ b/conf/council-lincolnshire_confirm.yml-example
@@ -12,3 +12,4 @@ reverse_status_mapping: {}
 attribute_descriptions: {}
 ignored_attribute_options: []
 default_site_code: ""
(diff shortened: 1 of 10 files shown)


MorayMySoc (approved):
Looks good to me

chrismytton:
@davea Looks good to me! Thanks for picking this up and getting it over the line 🏆

PR #231, opened 2022-11-08
[Southwark] Add Confirm integration

davea:


conf/council-southwark_confirm.yml-example
```diff
@@ -0,0 +1,11 @@
...
+username: ""
+password: ""
+tenant_id: ""
+server_timezone: ""
+cutoff_enquiry_date: ""
+service_whitelist: {}
+ignored_attributes: []
+forward_status_mapping: {}
+reverse_status_mapping: {}
+
```
dracos, line 11: Want to have the two new fields mentioned in this example?

davea:
Good idea 👍

dracos (commented):
Would be good to use POD rather than comments (sorry) :)

dracos (approved):
Thanks :)

PR #232, opened 2022-11-11
[Brent] Initial setup

dracos:
Think this is the skeleton of what's needed, barring additional information to come in about e.g. XML updates, special cases to implement.


chrismytton:
Just to note that I've done some very similar changes for Camden in https://github.com/mysociety/open311-adapter/pull/233, so there may be some overlap/conflicts when they both get merged into master.

conf/council-brent_symology.yml-example
```diff
@@ -1,4 +1,6 @@
 endpoint_url: ''
+endpoint_username: ''
+endpoint_password: ''
```
chrismytton, line 3: Is the endpoint username different from the `username: FMS` that we have for Bexley/Central Beds (that's used in `perllib/Open311/Endpoint/Integration/Symology.pm`, or are they the same? Just a bit wary of having two usernames floating around!

chrismytton (approved):
One question about possible confusion around multiple Symology usernames, but otherwise this looks good to go!

conf/council-brent_symology.yml-example
dracos: I'm afraid they are different, yes. One is the username for connecting/access to the endpoint, the other is the in-Symology username of the user making the request.

PR #233, opened 2022-11-14
[Camden] Symology integration - initial setup

chrismytton:
Setup Camden Symology integration. Doesn't handle updates currently as we need to write the integration to pull XML from the Symology SFTP server for that, which I will deal with separately from this PR.

This is similar to the Brent changes in https://github.com/mysociety/open311-adapter/pull/232, so this code might change if that gets merged into master first.

Fixes https://github.com/mysociety/societyworks/issues/3304


davea (approved):
🟢

PR #234, opened 2022-11-15
[Symology] Add request default config.

dracos:
This will let someone specify e.g. a default fallback Priority.
Could simplify the existing Central Beds config as well with this.


chrismytton (approved):
Nice! This makes it easier to set the ServiceCode globally for Camden as well, rather than doing it per-category.

PR #236, opened 2022-11-30
Allow any characters in received attributes.

dracos:
This will fix why `CREW NOTES` is not arriving at Bartec.
Fixes https://github.com/mysociety/societyworks/issues/3369


davea (approved):
This change is fine in itself, however I noticed when testing that we don't allow spaces in codes on the FMS side. This means we won't be able to edit the `Bulky collection` contact in the admin - don't know if that matters? It'll probably bite us at some point.

dracos:
Maybe, but I guess it can wait until then if it does, the attribute *does* at least have space in this case, so I guess passing it through like that does make sense for now.

PR #239, opened 2022-12-12
[Confirm] Add request_ignore_statuses config.

dracos:
This lets you specify a status code that should be ignored for fetched requests, but still used for updates.


davea (changes requested):
Looks OK but a test would be good please :)

PR #240, opened 2022-12-13
[Brent] Map title attribute to location

MorayMySoc:
FMS will be passing through the report title in an attribute so it can be mapped to the location for symology

https://github.com/mysociety/societyworks/issues/3407


dracos (approved):
Lovely :)

PR #242, opened 2022-12-16
[Echo] Initial changes for FMS reporting

dracos:
This lets us specify which Echo servers are waste and which aren't (for group/type), lets us have category grouping (wasn't needed for Waste), and lets us report without a UPRN (I’ve used USRN for Bromley, docs imply it should work only with lat/long but it did not for me).


MorayMySoc (approved):
Approved

PR #244, opened 2023-01-10
[Open311] Allow non-word characters in attributes from FMS admin

MorayMySoc:
Allows non-word characters to be passed through to endpoints as attributes.

Change made in response to Brent having a question mark in some of their names for extra questions

### DIFF
--- a/perllib/Open311/Endpoint.pm
+++ b/perllib/Open311/Endpoint.pm
@@ -459,7 +459,7 @@ sub POST_Service_Request_input_schema {
         $attributes{ $section }{ $key } = $def;
     }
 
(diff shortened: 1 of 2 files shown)


dracos (changes requested):
Having this in a test would be good, I think only needs a line like `'attribute[What is this?]' => 'starfish',` to around line 306 in t/open311/endpoint.t (which is where the any attributes code is already tested).

dracos (approved):

MorayMySoc:
Merged to master and staging, but forgot to push the rebase so automatic merge failed

PR #248, opened 2023-01-31
[Echo] Extra attribute configuration

dracos:
Plus couple of update bug fixes, missing required attribute, and documentation.

### DIFF
--- a/perllib/Open311/Endpoint/Integration/Echo.pm
+++ b/perllib/Open311/Endpoint/Integration/Echo.pm
@@ -1,3 +1,15 @@
+=head1 NAME
+
+Open311::Endpoint::Integration::Echo - An integration with the Echo backend
+
+=head1 SYNOPSIS
+
+This integration lets us post reports (as Events) and updates (as Event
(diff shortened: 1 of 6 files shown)


perllib/Open311/Endpoint/Integration/Echo.pm
```diff
@@ -66,6 +66,17 @@ incoming service codes don't match that actually in use).
 
 has service_mapping => ( is => 'ro', default => sub { {} } );
 
+=head2 service_extra_data
+
+This lets you list extra attributes that are required for a particular
```
MorayMySoc, line 71: The required in this comment may be confusing as they may be needed for that category, but the hidden fields are not required

MorayMySoc (approved):
Seems good to me. Minor query on description

perllib/Open311/Endpoint/Integration/Echo.pm
dracos: Have changed to "This lets you list extra attributes that should be added for a particular category"

PR #249, opened 2023-01-31
Set different comment on uploads if bulky collection

nephila-nacrea:
For https://github.com/mysociety/societyworks/issues/3410.

Peterborough want bulky goods collection photos to be marked with the comment 'Bulky waste photo' (see https://3.basecamp.com/4020879/buckets/26662378/todos/5608057275).

### DIFF
--- a/perllib/Open311/Endpoint/Integration/Bartec.pm
+++ b/perllib/Open311/Endpoint/Integration/Bartec.pm
@@ -211,12 +211,13 @@ sub post_service_request {
         $self->logger->warn("failed to attach note to ServiceRequest: " . $res->{ServiceCode} . " (FMS ID " . $args->{attributes}->{fixmystreet_id} . ")");
     };
 
-
     try {
-        if ( @{ $args->{media_url} }) {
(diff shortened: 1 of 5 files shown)


davea (changes requested):
Looks like a nice simple change, but please chuck a test in :)

perllib/Open311/Endpoint/Integration/Bartec.pm
```diff
@@ -212,8 +212,7 @@ sub post_service_request {
     };
 
     try {
-        $args->{service_name}
-            = $sr->{ServiceRequest}{ServiceType}{Name} // '';
+        $args->{service_name} = $service->service_name;
```
davea, line 215: We already have the category name here so seemed easier to use that rather than whatever Bartec returns (plus it makes the test easier as the XML doesn't include that value!)

davea:
@nephila-nacrea I've added a test, hopefully that looks OK.

nephila-nacrea:
> @nephila-nacrea I've added a test, hopefully that looks OK.

Thank you; that looks good to me - would you be able to approve the PR?

davea (approved):

PR #252, opened 2023-02-13
[SLWP] Store DD payment type separately.

dracos:
Fixes https://3.basecamp.com/4020879/buckets/26876928/todos/5827020769


MorayMySoc (approved):
Looks good to me.

PR #256, opened 2023-03-06
[Bartec] Upgrade to R16 API

davea:
Only a few renamed API methods, no substantive changes to logic or anything.


perllib/Integrations/Bartec.pm
```diff
@@ -213,9 +213,9 @@ sub _methods {
             namespace  => 'http://bartec-systems.com/',
             parameters => [],
         },
-        'ServiceRequests_Notes_Create' => {
+        'ServiceRequest_Note_Create' => {
```
dracos, line 216: The text file you link to says "E.g. Premises_Attributes_Create becomes Premises_Attribute_Create" - the first one remained plural in that example, so just checking this goes singular in both parts...

dracos (commented):
I assume it's fine, just the one query about one call name :)

perllib/Integrations/Bartec.pm
davea: Good question! I wondered the same, but it's correct according to the API: https://collectiveapi.bartec-systems.com/API-R1604/CollectiveAPI.asmx?op=ServiceRequest_Note_Create

dracos (approved):

PR #257, opened 2023-03-07
[Symology] Remove local XML files more than two days old

chrismytton:
This code was searching for files with ctime exactly two days ago, instead of two-or-more days ago.

The reason the code uses `-ctime +1` instead of `-ctime +2` is documented on the find(1) man page:

> When find figures out how many 24-hour periods ago the file was last accessed, any fractional part is ignored, so to match `-atime +1`, a file has to have been accessed at least two days ago.

(`-ctime` has the same behaviour as `-atime`)


chrismytton:
Closing in favour of https://github.com/mysociety/open311-adapter/pull/258.

PR #258, opened 2023-03-07
[Symology] Don't overwrite existing files retrieved from SFTP

chrismytton:
When moving the files retrieved from SFTP to their destination tell `mv` to skip any files that already exist.

This should fix the cleanup script, which currently isn't removing files because they're getting updated every time they're pulled.


dracos (approved):
Woo, hopefully this will work :)

PR #261, opened 2023-03-16
[Echo] Allow per-event type data_key_open311_map.

dracos:
For assisted collection name passthrough


MorayMySoc (approved):
Looks good to me.

PR #264, opened 2023-03-21
[Echo][Brent] Adds extra mandatory fields for missed collection

MorayMySoc:
Missed collections must check mandatory fields. Currently we can't see how to differentiate between bins and bags so check both fields for any missed collection although only one field is mandatory.

Relates to Brent WW go-live: Missed collection events https://3.basecamp.com/4020879/buckets/28996137/todolists/5888021261


dracos (approved):
looks good :)

PR #266, opened 2023-04-04
Support config-defined category and group aliases in Alloy integrations.

neprune:


chrismytton (approved):
Looks good! Nice to be able to have everything under the `service_whitelist` key, rather than needing a separate key for the groups.

dracos (commented):
Agree with Chris, this looks fine - one thing I wanted to check, does this work for the reports coming from FMS? I can't see anything (sorry if I've missed it) that would map back from the service code passed in by FMS to the alloy codes to then match in the find group/category code.

neprune:
@dracos yes - service code construction is unchanged so decomposing the service code for names to lookup in Alloy still works https://github.com/mysociety/open311-adapter/blob/d57850d649bd717b56ef4518bc5f965ef2da237e/perllib/Open311/Endpoint/Integration/UK/NorthumberlandAlloy.pm#L45

dracos:
Aha, great, thanks. Got confused between code/name :)

PR #276, opened 2023-05-23
[Alloy] Handle failed attachment downloads.

neprune:
Tested locally in addition to test case. Working as expected.


dracos (changes requested):
One issue with the success case (and thoughts on how to test that) is all :)

PR #277, opened 2023-05-25
Adding a report for Abavus integration

MorayMySoc:
This is the integration set up to send a report to Abavus according to the specification.

Still to be added is the updating a status from a report already created.


t/open311/endpoint/abavus.t
```diff
@@ -0,0 +1,132 @@
...
+use Open311::Endpoint::Service::UKCouncil;
+
+BEGIN { $ENV{TEST_MODE} = 1; }
+
+my $bucks_endpoint = Open311::Endpoint::Integration::Abavus::Dummy->new(
+    jurisdiction_id => 'bucks_abavus',
+    config_file => path(__FILE__)->sibling("abavus.yml")->stringify,
+    );
+
+my @services = $bucks_endpoint->services;
```
dracos, line 74: Is fine, would probably group these test in a few subtests (e.g. 'services', 'posting report'), just for scanning through

t/open311/endpoint/abavus.t
```diff
@@ -0,0 +1,132 @@
...
+use HTTP::Headers;
+use Test::More;
+use JSON::MaybeXS qw(encode_json decode_json);
+use URI;
+
+
+extends 'Integrations::Abavus';
+
+sub _build_config_file { path(__FILE__)->sibling("abavus.yml")->stringify };
+sub api_call {
```
dracos, line 14: This mocking works fine, but does mean none of the api_call code is being tested (e.g. setting the auth header might be worth checking). Could you have the same checking, but at a lower level, perhaps done as a mock of LWP::UserAgent's `request` maybe?

perllib/Open311/Endpoint/Integration/Abavus.pm
```diff
@@ -0,0 +1,247 @@
...
+=head2 service_list
+
+This is a mapping of Abavus services to use for categories populating FMS
+
+=cut
+
+has service_list => (
+    is => 'ro',
+    default => sub {
+        return {} if $ENV{TEST_MODE};
```
dracos, line 81: As you're mocking this in the test, do you need these guards, here and below?

t/open311/endpoint/abavus.yml
```diff
@@ -0,0 +1,64 @@
+{
```
dracos, line 1: I'd have this either as YAML, or valid JSON :)

perllib/Open311/Endpoint/Integration/Abavus.pm
```diff
@@ -0,0 +1,247 @@
...
+            my $name = $servicelist->{$code};
+            my %service = (
+                service_name => $name,
+                description => $name,
+                service_code => $code,
+                group => $group,
+            );
+            my $o311_service = $self->service_class->new(%service);
+
+            my $data = $self->service_extra_data->{$code};
```
dracos, line 148: extra_code looks good, I wonder if extra_data could be tweaked slightly to make it easier. I'd change type to datatype (and value 'dropdown' to 'singlevaluelist') and change choices to values, and I'd make it a list rather than a hash, because question order might be important (presumably sorting won't always be correct). So I'd have:

```
  "service_extra_data": {
      DISPLAY_ISSUE_17821_C: [
        {
         code: "DISPLAY_ISSUE_TYPE_646538_I",
          description: "Incorrect or broken",
          required: 1,
          datatype: 'singlevaluelist',
          values: {
            DISPLAY_916276_A: Incorrect,
            DISPLAY_916277_A: Broken
          }
        },
        {
          code: "DISPLAY_CONTACT_646539_I",
          description: "May we contact you?",
          type: 'text',
          required: 0,
        }
      },
...
```
And then you can loop over and use each hash directly as %params, no need to map anything across, it's already what you want.

dracos (changes requested):
Looking good, some suggestions :)

davea (approved):
Looks good! One debugging line to remove, but 👍

PR #279, opened 2023-06-09
Abavus update read fields and add photos field

MorayMySoc:
Checks against the 'result' rather than the 'id' - there should be one result for creating a serviceRequest and 0 for failure. Seems more solid than the 'id' returning -1 which I couldn't see documented.

Also changes fields read to correct Abavus return field names.

Adds in a photo field to send the urls of photos on the report.

Also adds a test for the photo field and a check that all fields are being tested as currently there could be a silent failure if checking fields breaks down.

### DIFF
--- a/t/open311/endpoint/abavus.t
+++ b/t/open311/endpoint/abavus.t
@@ -11,6 +11,8 @@ use URI;
 extends 'Integrations::Abavus';
 
 my $lwp = Test::MockModule->new('LWP::UserAgent');
+my $lwp_counter = 0;
+
 $lwp->mock(request => sub {
     my ($ua, $req) = @_;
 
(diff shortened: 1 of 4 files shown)


t/open311/endpoint/abavus.t
```diff
@@ -31,11 +34,12 @@ $lwp->mock(request => sub {
         is $content->{serviceRequest}->{form}->{code}, 'ABANDONED_17821_C';
         is $content->{serviceRequest}->{location}->{latitude}, '50';
         is $content->{serviceRequest}->{location}->{longitude}, '0.1';
-        return HTTP::Response->new(200, 'OK', [], encode_json({"id" => 1}));
+        return HTTP::Response->new(200, 'OK', [], encode_json({"result" => 1, "id" => 1}));
```
dracos, line 37: This adding of `result` doesn't appear to be to do with this commit, but the previous one?

dracos (changes requested):
One thing in the wrong commit I think is all :)

dracos (approved):

PR #285, opened 2023-07-04
Upgrade carton to v1.0.34, run a fresh install and only test perl 5.30+.

neprune:


t/open311/endpoint/hackney_alloy.t
```diff
@@ -105,7 +105,6 @@ subtest "create basic problem" => sub {
         'attribute[description]' => 'description',
         'attribute[title]' => 'title',
         'attribute[report_url]' => 'http://localhost/123',
-        'attribute[asset_resource_id]' => undef,
```
neprune, line 108: For reasons unknown, after the fresh install, this line resulted in `attribute[asset_resource_id]` having a value of 1 in the arguments passed to `post_service_request`. Probably a bug introduced in a newer `Web::Simple`.

dracos: Latest version of URI (5.19) changed "Form parameters without values are now represented by undef" so it becomes a keyword not a parameter and so gets set to 1 as you found. I think not having it is fine.

dracos (approved):
Looks good, good to update these things every so often, will need a bit of thought about deploying.

neprune:
Not sure why this wasn't picked up in GH, but has been merged to master.

PR #287, opened 2023-07-11
[Brent] ATAK integration

chrismytton:
Setup Brent ATAK integration for creating issues in ATAK.

Fixes https://github.com/mysociety/societyworks/issues/3804

### Related FixMyStreet pull request

- https://github.com/mysociety/fixmystreet/pull/4551


dracos (approved):
Code looks good, assuming it's putting the right data in the right places :)

perllib/Integrations/ATAK.pm
```diff
@@ -0,0 +1,142 @@
...
+Returns the JSON response.
+
+=cut
+
+sub post {
+    my ($self, $path, $data) = @_;
+
+    my $url = $self->config->{api_url} . $path;
+    $self->logger->debug("[ATAK] Request URL: $url");
+    $self->logger->debug("[ATAK] Request data: " . encode_json($data));
```
chrismytton, line 105: This line is logging the base64 encoded images to the logfile, which isn't ideal. Need to filter the attachments out before logging or similar.

chrismytton:
Have pushed a fixup commit which converts the lat/long to numbers before sending to ATAK, and also makes a logging line a bit clearer.

chrismytton:
Have pushed a fixup to my fixup, which _actually_ makes the coordinates into numbers in the JSON.

Also pushed another fixup to set the mime-type of the image dynamically based on the image response, rather than always using JPEG.

perllib/Integrations/ATAK.pm
dracos: Rather than log here, perhaps log just after `$issue` is created in post_service_request? Then the token call isn't being logged all the time as well.

dracos (commented):
One comment about your comment on logging is all :)

PR #288, opened 2023-07-16
[Central Beds] Use SinglePoint to find the nearest address.

neprune:
Closes https://github.com/mysociety/societyworks/issues/3749


t/geocode/files/singlepoint/spatial_radial_search_by_easting_and_northing_response.xml
```diff
@@ -0,0 +1,3739 @@
+<?xml version="1.0" encoding="utf-8"?>
```
dracos, line 1: Could this test data perhaps be a bit smaller? :)

dracos (changes requested):
Again, looks good, can we shrink the test data a bit? Plus reusing XML::Simple as discussed, hopefully not a big change.

dracos (approved):
Thanks, glad it wasn't a big thing, looks good :)

PR #289, opened 2023-07-17
[Bexley] Update Symology status mapping

davea:
For FD-3182

### DIFF
--- a/perllib/Open311/Endpoint/Integration/UK/Bexley/Symology.pm
+++ b/perllib/Open311/Endpoint/Integration/UK/Bexley/Symology.pm
@@ -72,7 +72,7 @@ sub _update_status {
             'fixed'
         } elsif ($action_due =~ /^[NS][1-6]$/) {
             'in_progress'
-        } elsif ($action_due =~ /^IR|REH|RES|RET|RP|RPOS|RT|RWT$/) {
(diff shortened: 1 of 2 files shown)


perllib/Open311/Endpoint/Integration/UK/Bexley/Symology.pm
```diff
@@ -72,7 +72,7 @@ sub _update_status {
             'fixed'
         } elsif ($action_due =~ /^[NS][1-6]$/) {
             'in_progress'
-        } elsif ($action_due =~ /^IR|REH|RES|RET|RP|RPOS|RT|RWT$/) {
+        } elsif ($action_due =~ /^IR|RBC|REH|RES|RET|RP|RPL|RPOS|RSW|RT|RWT$/) {
```
dracos, line 75: Not sure if this is deliberate, but the lack of brackets (before) means this matched anything starting IR, anything ending RWT, plus anything containing the rest (so e.g. RPL already matched the RP one). Presumably it should have brackets to match e.g. the investigating one a few down.

dracos (changes requested):
What you've done is fine, but I think (assume) the existing code had a bug we should probably fix.

perllib/Open311/Endpoint/Integration/UK/Bexley/Symology.pm
davea: Good spot! Have added brackets.

dracos (approved):
Woo

PR #290, opened 2023-07-18
[Alloy] Cache reconstruct calls.

dracos:


perllib/Open311/Endpoint/Integration/AlloyV2.pm
```diff
@@ -1290,4 +1292,27 @@ sub _find_group_code {
...
 }
 
+sub call_reconstruct {
+    my ($self, $id, $date) = @_;
+
+    my ($dir, $file);
+    if ($dir = $self->update_store) {
+        $file = path($dir)->child("$id-$date.json");
+        if ($file->exists) {
+            return decode_json($file->slurp_raw);
```
davea, line 1302: This returns something different to L1315, think it needs to be this otherwise it'll crash:

```suggestion
            return decode_json($file->slurp_raw)->{item};
```

(Alternatively guess we could save a few bytes on disk by changing L1312 to just store `$resource->{item}`)

davea (changes requested):
There's a crash in there, not sure how the tests didn't pick it up though.

perllib/Open311/Endpoint/Integration/AlloyV2.pm
dracos: Test never calls the same thing twice, hence why it didn't crash, thanks for spotting, will try and make it do so.

dracos: I think I do want to store the whole thing as it includes metadata that theoretically might be useful (though I guess could always be looked up).

PR #293, opened 2023-08-08
Glos confirm integration post testing fixes

nephila-nacrea:
For the to-dos in https://3.basecamp.com/4020879/buckets/32187059/todolists/5997794619.


davea (commented):
This looks fine, but I wonder if it would be better to do the string manipulation in `Gloucestershire::open311_extra_data_include` in FMS, given that already exists? Doing it here in open311-adapter means that there's Gloucestershire-specific functionality in two places rather than one.

davea (approved):

PR #294, opened 2023-08-09
[Brent] [ATAK] Add support for fetching updates.

neprune:
https://github.com/mysociety/societyworks/issues/3805

I've belatedly realised I've over-engineered this since we're only expecting a single update to a terminal status - had a bit of Maslow's hammer moment having done something similar for Jadu.

I think it's fine to keep as is, as it does the job and opens the door for tracking intermediary states or changes from terminal states, if that's ever something we want, but happy to make changes if you'd prefer we shed the complexity.


t/open311/endpoint/brent.t
```diff
@@ -50,15 +51,50 @@ $brent_integ->mock(config => sub {
...
+        issue_status_tracking_max_age_days => 365,
+        update_storage_max_age_days => 365,
+        atak_status_to_fms_status => {
+            "Closed - Completed" => "fixed",
+            "Closed - Out of scope" => "no_further_action",
+            "Closed - Not found" => "closed",
+            "Closed - Passed to Brent" => "internal_referral",
+        },
+    }
+}
```
neprune, line 89: Moving the test config into code just so I can stick the tempfile handles in.

chrismytton (approved):
Seems to work really nicely!

PR #298, opened 2023-08-20
[Echo] Allow PostEvent GUID to be set by Open311

davea:
For e.g. bulky collections the same GUID used by FMS in the `ReserveAvailableSlotsForEvent` call must be used for subsequent `PostEvent` call.

For https://github.com/mysociety/societyworks/issues/3758

### DIFF
--- a/perllib/Integrations/Echo.pm
+++ b/perllib/Integrations/Echo.pm
@@ -169,6 +169,7 @@ sub PostEvent {
         );
     }
     my $data = ixhash(
+        $args->{guid} ? (Guid => $args->{guid}) : (),
         $args->{data} ? (Data => extensible_data($args->{data})) : (),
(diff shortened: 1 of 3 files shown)


perllib/Integrations/Echo.pm
```diff
@@ -170,6 +170,7 @@ sub PostEvent {
     }
     my $data = ixhash(
         $args->{data} ? (Data => extensible_data($args->{data})) : (),
+        $args->{guid} ? (GUID => $args->{guid}) : (),
```
dracos, line 173: This doesn't work here, the event is created but the guid provided is not used.
The line needs to come before Data and must be `Guid`.

dracos (changes requested):

PR #299, opened 2023-08-22
[ATAK] Use the ATAK ID for service_request_id

chrismytton:
This was using the FixMyStreet external ID as the service_request_id, which meant it wasn't getting matched when the updates were pulled into FMS.

Use the ATAK ID rather than the FMS reference as the service_request_id.

Follows on from #294


dracos (approved):
Looks good, assume it fixes the issue involved :)

PR #300, opened 2023-08-25
[Echo] Allow PostEvent task reservation to be passed in.

dracos:
Similar to #298 but for the chosen reservation slot.


MorayMySoc (approved):
Looks good to me

PR #306, opened 2023-09-14
[Alloy] Improve service code usage.

dracos:
Assign unique codes to categories with an alias, and spot categories with the same name in multiple groups; use the group passed in directly where necessary (Northumberland). This should allow the use of categories with the same name in multiple groups without having to have different names (meaning most aliases can be removed from configuration) and allow two different categories to map to the same Alloy category.

Pairs with origin/alloy-send-group in FMS repo to send the group to here.

Should help resolve FD-3318 for Bucks and improve/simplify category display for Northumberland.


neprune (approved):
Nice! 🎉

PR #309, opened 2023-10-17
[Bromley] [Echo] Setup bulky cancellations.

neprune:
https://github.com/mysociety/societyworks/issues/3895


dracos (approved):
Looks good :) I guess with the next one, if it's similar to this/K&S (as it was not for Brent), we might want to factor out some form of config to do this, for now this works fine.

dracos:
This was rebased/merged as 743db5a, remember to push the branch before pushing the merge so it can auto-close :)

PR #314, opened 2023-11-14
[Bartec] Encode email in case of special characters

MorayMySoc:
Add encoding to email addresses in case of special characters in email addresses prompted by an email with an ampersand in it


dracos (approved):
Woo

PR #315, opened 2023-11-21
[Northumberland] Fetch assigned-to users

nephila-nacrea:
Note: also need to deploy changes to northumberland alloy config on mysociety-servers.

This is the open311-adapter portion of https://github.com/mysociety/societyworks/issues/3978. We look for assigned-to users in each update from Alloy and pass email and name through to FMS via the 'extras' field.

[skip changelog]


perllib/Open311/Endpoint/Integration/AlloyV2.pm
```diff
@@ -633,13 +636,39 @@ sub _get_inspection_updates {
                 updated_datetime => $update_dt,
             );
 
+            my $assigned_to_user = {};
```
dracos, line 639: Don't think this line is needed, could be `my $assigned_to_user` inside the if.

perllib/Open311/Endpoint/Integration/AlloyV2.pm
```diff
@@ -633,13 +636,39 @@ sub _get_inspection_updates {
...
+            if ( my $assigned_to_user_id
+                = $attributes->{ $mapping->{assigned_to_user} // '' }[0] )
+            {
+                $assigned_to_user
```
dracos, line 643: This block is fine, wondered if you'd like to rejig it to something like:

```
my $assigned_to_user = $assigned_to_users->{$assigned_to_user_id} ||= do {
    # There is a possibility the assigned-to user is not already
    # in the $assigned_to_users hash; do another lookup if so
    $self->get_assigned_to_users($resource)->{$assigned_to_user_id};
};
```
which then handles the checking/reassignment in a neater block.

dracos (changes requested):
Looks good, all clear to read 👍  Couple of small suggestions to the block is all :)

perllib/Open311/Endpoint/Integration/AlloyV2.pm
nephila-nacrea: Ooh nice and pithy, I wasn't that familiar with using do{} in this kind of way :-)

dracos (approved):
🧑‍🚀

PR #317, opened 2023-11-27
[Symology] Use a top-level TMPDIR when fetching files

sagepe:
The trap to remove the TMPDIR being set-up within the loop doesn't work as expected as it gets updated with each iteration and leaves directories behind.

This sets up a top-level TMPDIR and sets a single global trap to clean that up on exit.

### DIFF
--- a/bin/symology/fetch_files
+++ b/bin/symology/fetch_files
@@ -4,7 +4,14 @@ set -eu
 
 source /data/mysociety/shlib/deployfns
 
+TMPDIR=$(mktemp -d) || exit 1
+trap 'rm -rf "$TMPDIR"' EXIT
+cd $TMPDIR
+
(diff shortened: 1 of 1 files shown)


bin/symology/fetch_files
```diff
@@ -4,22 +4,30 @@ set -eu
...
     echo "get $OPTION_updates_sftp__dir data" | sshpass -e sftp -oBatchMode=no -r -b - $OPTION_updates_sftp__username@$OPTION_updates_sftp__host >/dev/null
     if ls -A1q "data" | grep -q .; then
         mv -n data/* $OPTION_updates_sftp__out
     fi
 
+    cd ..
+
 done
+
+cd
```
dracos, line 32: Do you need these lines at the end?

PR #318, opened 2023-12-14
[Brent][ATAK] Set external status code for closed reports

chrismytton:
We were already setting the FMS status code correctly, but were missing the external status code for reports that were automatically marked as closed, which was causing some warnings.

### DIFF
--- a/t/open311/endpoint/brent.t
+++ b/t/open311/endpoint/brent.t
@@ -88,6 +88,7 @@ sub atak_config {
             "Closed - Passed to Brent" => "internal_referral",
         },
         fixed_status => "Closed - Completed",
+        closed_status => "Closed - Not found",
     }
 }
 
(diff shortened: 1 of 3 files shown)


t/open311/endpoint/brent.t
```diff
@@ -927,7 +939,7 @@ subtest "GET ATAK service request updates OK" => sub {
                     "task_d_planned": "2023-08-02T01:00:00Z",
                     "task_d_completed": "2023-08-02T02:00:00Z",
                     "task_d_approved": "2023-08-04T03:00:00Z",
-                    "task_p_id": "128"
+                    "task_p_id": "127"
```
chrismytton, line 942: This isn't directly related to this change, I just noticed that I hadn't updated this ID when I made the previous changes in https://github.com/mysociety/open311-adapter/pull/310.

dracos (approved):
👟

PR #319, opened 2023-12-18
[Glos] Subscribe to jobs from Confirm

nephila-nacrea:
In addition to the usual enquiries.

Fetching jobs requires us to use GraphQL (see https://help.dudesolutions.com/Content/PDF/Confirm/v21.10/confirm-v21-10-web-api-specification.pdf), though Confirm's implementation of GraphQL seems to lack features such as [support for variables](https://graphql.org/learn/queries/#variables), so I have had to use string interpolation.

For https://github.com/mysociety/societyworks/issues/4029.


perllib/Open311/Endpoint.pm
```diff
@@ -366,7 +366,8 @@ sub GET_Service_List {
             keywords => (join ',' => @{ $service->keywords } ),
             metadata => $self->format_boolean( $service->has_attributes ),
             @{$service->groups} ? (groups => $service->groups) : (group => $service->group),
-            map { $_ => $service->$_ } 
+            ( is_confirm_job => $self->format_boolean(1) ) x !!$service->is_confirm_job,
```
dracos, line 369: One note as I was looking at this PR - I would probably not create a new top-level entry for a service, but instead add an "is_confirm_job" to the keywords for the service (in perllib/Open311/Endpoint/Integration/Confirm.pm), that seems a more generic place to put it. If I've missed a reason to do it this way instead, please do say :)

perllib/Open311/Endpoint.pm
```diff
@@ -366,7 +366,7 @@ sub GET_Service_List {
             keywords => (join ',' => @{ $service->keywords } ),
             metadata => $self->format_boolean( $service->has_attributes ),
             @{$service->groups} ? (groups => $service->groups) : (group => $service->group),
-            map { $_ => $service->$_ } 
+            map { $_ => $service->$_ }
```
nephila-nacrea, line 369: Pointless change, I should remove it!

davea: Always good to tidy these things up if you spot them!

perllib/Open311/Endpoint/Integration/Confirm.pm
```diff
@@ -759,6 +882,92 @@ sub get_service_requests {
         push @requests, $request;
     }
 
+    # Jobs
+
+    # We try to follow a similar pattern to Enquiries above; we find updated
+    # jobs by first grabbing the status logs for the given time period, then
+    # look up the details for each job associated with those status logs.
```
davea, line 889: Is doing it this way necessary here? The reason we do it that way for Enquiries is because the Enquiries API doesn't allow us to query for Enquiries raised within a specific timeframe. The GraphQL API does let us do this for jobs, however, so unless there's a reason I've overlooked it might be better to just do a single API call here if we can.

```graphql
{
  jobs(
    filter: {
      entryDate: {
        greaterThan: "2024-01-23T09:00:00"
        lessThan: "2024-01-23T10:30:00"
      }
    }
  ) @_size_1000 {
    description
    entryDate
    jobType @_size_1000 {
      code
    }
    jobNumber
  }
}

```

davea (commented):
This works really nicely! I've got one question about the approach taken but overall this is great.

perllib/Open311/Endpoint.pm
nephila-nacrea: (As in, keep this change, or remove it? This was accidental 😅 as Endpoint.pm seems to have some funny whitespace which my editor changes to something more standard when I save the file. Ended up removing changes I'd originally made to Endpoint.pm but this funny little whitespace stayed.
Eh, probably not worth fussing over, if it's here now, it's here...)

perllib/Open311/Endpoint/Integration/Confirm.pm
nephila-nacrea: That looks way simpler and was probably the way I was doing it originally, I overthought things and thought I ought to be following the enquiries example 😖 'the Enquiries API doesn't allow us to query for Enquiries raised within a specific timeframe' - I hadn't realised this!

davea (approved):
Looks good!

PR #330, opened 2024-02-29
[Alloy] Strip multiple extensions from the service code.

dracos:
This is to allow two subcategories in the same category to go to the same destination category in Alloy, e.g.

Road and street signs:
  Sign light not working: { "alias": "Sign with a light (blue)" },
  Sign light not working_1: { "alias": "Sign with a light (other)" },

It was fine if they were in different top-level categories, but impossible if they were in the same. This is a bit of a workaround, but seemed to be okay. For FD-3900.


neprune (approved):
nice, lgtm

PR #333, opened 2024-04-11
[Bexley][Whitespace] Use custom message for worksheet

chrismytton:
Instead of passing through the FMS description Bexley want to have details about whether a property is assisted and the location of the containers in the message that goes to Whitespace.

For https://3.basecamp.com/4020879/buckets/35109031/todos/7265317231

Part of https://github.com/mysociety/societyworks/issues/4206


chrismytton:
@davea Have added a (hopefully!) better test now, so this is ready for review 🙂

davea (approved):
🎉

PR #335, opened 2024-04-12
[Bromley][WW] Truncate text to maximum field length

MorayMySoc:
Notes text should not exceed maximum length for notes field.

https://github.com/mysociety/societyworks/issues/4232
https://3.basecamp.com/4020879/buckets/30553621/todos/7187142360#__recording_7288097933


lizettal:
@dracos are you able to prioritise reviewing this so we can get this across to Bromley as then I think everything is with them for testing.

dracos (approved):

PR #337, opened 2024-05-08
[Bexley][Whitespace] Remove service mapping script

chrismytton:
Bexley have provided us with a manually created mapping between service items and services, meaning this script is no longer needed.

Removing the script also means that we can't accidentally overwrite the service mapping they provided in the future.

Relates to the [Wrong worksheet created](https://3.basecamp.com/4020879/buckets/35109031/todos/7265317231) todo on Basecamp.


dracos (approved):
Fine, though I do like the quasi-documentation of the script - could the example file perhaps have a couple of examples as well?

chrismytton:
Oops, I saw the approval come through as a Slack notification but failed to actually check for comments 🤦 

Yes, I'll add a couple of examples to the example config.

PR #348, opened 2024-07-12
[Echo] Strip suffixes defined by _

MorayMySoc:
If we have used a suffix in the categories of FMS to allow us to use one Event Type with separate category names, we strip the suffix before sending to Echo so the report is sent to the base Event type

https://github.com/mysociety/societyworks/issues/4365

### DIFF
--- a/perllib/Open311/Endpoint/Integration/Echo.pm
+++ b/perllib/Open311/Endpoint/Integration/Echo.pm
@@ -422,7 +422,11 @@ sub process_service_request_args {
     my $self = shift;
     my $args = shift;
 
-    my $event_type = $args->{service_code};
+    # Strip off any extra data added as a suffix in FMS config
+    # to reuse event type with multiple category names
+    # - ie 2951, 2951_1 and 2951_2 should be posted as 2951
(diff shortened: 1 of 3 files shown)


MorayMySoc:
I don't think Symology or ATAK should be affected as you said - couldn't see how if so.

perllib/Integrations/Echo.pm
```diff
@@ -170,12 +170,17 @@ sub PostEvent {
             ),
         );
     }
+    # Strip off any extra data added as a suffix in FMS config
+    # to reuse event type with multiple category names
+    # - ie 2951, 2951_1 and 2951_2 should be posted as 2951
+    # as that is the actual EventTypeId
+    my ($event_type) = split /_/, $args->{event_type};
```
dracos, line 177: I think this is happening too late in the process? We'd want to be able to e.g. use the actual event type ID in `_get_data_value` for attribute lookups, not the 'duplicate' one.

`process_service_request_args` when you first get the service_code coming in would be better, I think.

dracos (changes requested):

dracos (approved):

PR #352, opened 2024-08-02
[Camden] Confirm trees location field contains nearest address data

MorayMySoc:
https://3.basecamp.com/4020879/buckets/36530864/todos/7607947511#__recording_7668470194


t/open311/endpoint/camden.t
```diff
@@ -291,4 +291,69 @@ subtest "GET updates OK" => sub {
     ok 1;
 };
 
+
+my $confirm_integ = Test::MockModule->new('Integrations::Confirm');
```
dracos, line 295: It looks like other Multi have one file per backend, might be clearer to do the same here? Hopefully is just a copy and paste.

dracos (changes requested):
Thanks for adding the test - only suggestion is moving it to its own file :)

dracos (approved):

PR #353, opened 2024-08-06
[Surrey] Expect serviceId from integration

MorayMySoc:
serviceId will be received from new version of Boomi integration so is not hard-coded

https://3.basecamp.com/4020879/buckets/36546415/todos/7648793252#__recording_7674131654 https://mysociety.slack.com/archives/C06SKSH5KQT/p1722624461300599?thread_ts=1722616474.424329&cid=C06SKSH5KQT


dracos (approved):
Haven't tested explicitly, but matches the requested change, yep

PR #354, opened 2024-08-06
[Surrey] Adds externalId for jobs and reports

MorayMySoc:
Surrey will be sending an externalId specifically to be used for imported jobs and reports

https://3.basecamp.com/4020879/buckets/36546415/todos/7672617616#__recording_7683887161 https://mysociety.slack.com/archives/C06SKSH5KQT/p1722948392215939


dracos (changes requested):
From the spec doc you link to here, externalId would be on fmsReport, not inside confirmEnquiry or confirmJob.

dracos (approved):

PR #356, opened 2024-08-14
[Surrey][Boomi] Fetch job updates

MorayMySoc:
Updates for Jobs are a separate integration id to
updates for Enquiries. Allow array of integration ids and receive Jobs updates and convert them to an FMS update.

https://3.basecamp.com/4020879/buckets/36546415/todos/7696104321


davea (approved):
Haven't been able to get the live API to return any job updates; but assuming it does so in line with the API spec then this looks good 👍

PR #361, opened 2024-09-17
[Surrey] Don’t fetch & upload photos from media_url

davea:
Photos should only be uploaded if the photo content was POSTed from FMS.


davea:
Yes, we can either provide the `base64` _or_ `url`.

dracos (approved):

PR #364, opened 2024-09-20
[Surrey] Don’t pass blank updates to Boomi

davea:


perllib/Open311/Endpoint/Integration/Boomi.pm
```diff
@@ -294,7 +294,11 @@ sub post_service_request_update {
     my $hash = substr(md5_hex(join('', @parts)), 0, 8);
     my $update_id = $args->{service_request_id} . "_" . $hash;
 
-    my $service_request_id = $self->boomi->upsertHighwaysTicket($ticket);
+    if ($args->{description}) {
+        # State changes on FMS may create updates with empty descriptions, which
+        # we don't want to send to Boomi as it will return an error.
+        my $service_request_id = $self->boomi->upsertHighwaysTicket($ticket);
```
dracos, line 300: This is setting a var that isn't used - presumably it's not needed (and wasn't before either), it uses args->{service_request_id}?

dracos (changes requested):

dracos (approved):

PR #366, opened 2024-10-15
[Confirm] Allow ignoring of enquiries in certain status

davea:
Fetched Enquiries will be skipped if their status
maps to IGNORE in reverse_status_mapping.

Matches behaviour of update fetching in 42f75f6b.


dracos (approved):
Assuming we're happy not to get a report at all if it's in this state

PR #369, opened 2024-10-15
[Bexley][WW] Create worksheets for container requests

nephila-nacrea:
For https://github.com/mysociety/societyworks/issues/4575 & https://github.com/mysociety/societyworks/issues/4576.

Also have config changes in a `bexley-ww-container-requests` branch in mysociety-servers.


nephila-nacrea:
Also have config changes in a `bexley-ww-container-requests` branch in mysociety-servers.

dracos (approved):
Looks good, I guess in future perhaps we want to match the `service_params` to the `service_code` and then you could look it up by that, e.g. $service_params->{$params->{service_code} . "_service_id"} or something like that, but not needed now :)

PR #376, opened 2024-12-03
[N/land] Pass category updates to/from Alloy

nephila-nacrea:
Fixes https://github.com/mysociety/societyworks/issues/4648.


dracos (changes requested):
My concern here is we've just cut down the API calls used, and this adds two per update, so I'd be worried of hitting the rate limit. I think we can change the `fetch_updated_resources` call to have a join to the category/group so that we get that information as part of the original call, which I think we should look into doing here.

The code itself looks good :)

dracos (approved):
Great :)

PR #377, opened 2024-12-19
[Alloy] Add CLI tools for inspecting and interacting with the Alloy API

chrismytton:
Adds a couple of command line tools for inspecting designs in an Alloy instance, and interacting with the Alloy API via the methods in the AlloyV2 module.

- `bin/alloy/inspect-designs` - Lets you view one or all designs in a given Alloy instance and shows attributes, their types and whether they're required in a text format.
- `bin/alloy/invoke` - Allows you to call methods in the AlloyV2 module from the command line and pass parameters in via JSON. Inspired by the `bin/fixmystreet.com/call-wasteworks-backend` tool in FMS.

### DIFF
--- a/bin/alloy/inspect-designs
+++ b/bin/alloy/inspect-designs
@@ -1,0 +1,129 @@
+#!/usr/bin/env perl
+
+=head1 NAME
+
+inspect-designs - Inspect and validate Alloy V2 API designs and configurations
+
+=head1 SYNOPSIS
+
+    # Inspect a single design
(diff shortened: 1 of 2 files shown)


chrismytton:
@MorayMySoc These might be useful for Bristol. Once you've got an API key you should be able to put it into a config file and then use the `inspect-designs` script to get details about the design, what the attributes are, attribute types and whether they're required or not.

And then when you're working on the integration the `invoke` script might come in handy when you need to see what the code is doing/seeing.

MorayMySoc (approved):
This works well for me - thank you!

PR #378, opened 2025-01-08
[Bristol] Convert Bristol to a Multi

MorayMySoc:
Bristol has a passthrough which needs to continue
to work alongside the new Alloy integration.

https://github.com/mysociety/societyworks/issues/4683


perllib/Open311/Endpoint/Integration/UK/Bristol.pm
```diff
@@ -0,0 +1,31 @@
+=head1 NAME
+
+Open311::Endpoint::Integration::UK::Bristol - Bristol integration set-up
+
+=head1 SYNOPSIS
+
+Bristol manage their own Confirm integration but have a managed Alloy integration, so is
```
chrismytton, line 7: I wonder if it might be clearer to simply state that Bristol have their own Open311 server, rather than specifically mentioning Confirm. What they do behind their Open311 endpoint might change at some point, but that doesn't matter to us.

chrismytton (changes requested):
Looks good so far! Couple of very minor comments.

chrismytton (approved):

MorayMySoc:
Sorry - something went odd when resetting the branch to merge to - from master to bristol, so not actually 72 changed files

dracos:
Noting here that the Passthrough doesn't actually appear to work at present, as the Bristol endpoint is returning an undefined datatype.

PR #383, opened 2025-01-23
[Bristol] Sending of reports

MorayMySoc:
Implements Bristol code for sending of reports from FMS to Alloy.

https://mysociety1.sharepoint.com/:x:/s/SocietyWorks/EfBblFantSxIjHxdQruMjigB0YiQ57i6rOwZLvuH2x7vug?rtime=6gwCKKk73Ug https://3.basecamp.com/4020879/buckets/39557940/uploads/8117287728

### DIFF
--- a/perllib/Open311/Endpoint/Integration/AlloyV2.pm
+++ b/perllib/Open311/Endpoint/Integration/AlloyV2.pm
@@ -208,11 +208,17 @@ sub post_service_request {
     $category =~ s/(_\d+)+$//;
     $category =~ s/_/ /g;
     $args->{service_code_alloy} = $category;
+    my $rfs_design;
+    if ($self->can('pick_design')) {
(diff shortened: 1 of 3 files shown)


MorayMySoc:
I'm going to do tests separately, but wanted to get what I had done visible...hopefully not now too over-engineered!

MorayMySoc:
This still also requires photo attachments

dracos (changes requested):
Looking good, one bug I think and couple of tiny suggestions is all.

MorayMySoc:
Realised that the 'JobType' wasn't quite right as is purely invented in the script and not an actual attribute...so made changes there too.

dracos (commented):
Code looks good, sorry for not spotting that jobtype thing, but the lack of tests of the Bristol specific code is a concern, would be good to rectify that soon.

PR #386, opened 2025-01-31
Ignore missing config files in development server.

dracos:


MorayMySoc (approved):
Very nice - thank you

PR #387, opened 2025-02-03
[Bristol] sending of updates

dracos:


MorayMySoc (approved):
This seems to work with https://github.com/mysociety/fixmystreet/pull/5352 at the FMS end

PR #389, opened 2025-02-06
[Bristol] Add test file and tests for sending reports

MorayMySoc:
Adds the test file initiated with tests for sending reports

https://github.com/mysociety/societyworks/issues/4770


dracos (changes requested):
Looks good, just couple of tiny things

dracos (approved):

PR #390, opened 2025-02-07
[Bristol] Default parent to road network item.

dracos:


perllib/Open311/Endpoint/Integration/UK/Bristol/Alloy.pm
```diff
@@ -108,4 +97,36 @@ sub process_attributes {
...
+    if (!%$parents) {
+        # Default to the road
+        my $road = $self->_get_usrn($args);
+        $parents->{attributes_defectsAssignableDefects} = [ $road->{itemId} ];
+    }
+
+    return $parents;
+}
+
+sub _get_usrn {
```
MorayMySoc, line 114: Maybe this should be called _get_road/_get_street ?

dracos: Oh yes, it's not getting a USRN is it!

PR #391, opened 2025-02-08
[Alloy] Allow using a join to check if defects are linked.

neprune:
Rather than separate per-item parent calls which are hitting the rate limit.
https://github.com/mysociety/societyworks/issues/4682.

On staging. Configuration commit on servers is `7c0c1a46f59d0bec7641e42279f4675a88c39448`. Live configuration will just be case of removing the staging guards - the values should be the same.

### DIFF
--- a/perllib/Integrations/AlloyV2.pm
+++ b/perllib/Integrations/AlloyV2.pm
@@ -168,7 +168,7 @@ sub design_attributes_to_hash {
 sub search {
     my ($self, $body_base, $skip_count) = @_;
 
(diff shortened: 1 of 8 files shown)


dracos (approved):
Looks good; just to check, the rfs_specific_attribute is just so we have something within the wanted design to join against, so it fetches the right joined item.

neprune:
> Looks good; just to check, the rfs_specific_attribute is just so we have something within the wanted design to join against, so it fetches the right joined item.

Yep, that's it.

PR #392, opened 2025-02-10
[Alloy] Add list/search of designs.

dracos:
And pretty print tables (add new module so that it can wrap tables to terminal width).


chrismytton (approved):
Nice! `Term::Table` is a big improvement for the output formatting 👍

PR #395, opened 2025-02-18
[Bristol][Alloy] Adds extra street cleansing categories

MorayMySoc:
Moving from having all street cleansing categories under one group, to having street cleansing category
in another group.

Specifically, the bike category wants to fit into the main FMS bike category.

### DIFF
--- a/perllib/Open311/Endpoint/Integration/UK/Bristol/Alloy.pm
+++ b/perllib/Open311/Endpoint/Integration/UK/Bristol/Alloy.pm
@@ -33,7 +33,12 @@ sub process_attributes {
 
     my $attributes_values = $self->config->{'request_attribute_to_values'};
     my $code = $args->{service_code_alloy};
-    if (grep { $code =~ /$_/ } keys %{$self->config->{service_whitelist}->{'SC-Street Cleansing'}}) {
+    if (grep { $code =~ /$_/ } keys %{
+            {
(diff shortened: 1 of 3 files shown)


perllib/Open311/Endpoint/Integration/UK/Bristol/Alloy.pm
```diff
@@ -33,7 +33,12 @@ sub process_attributes {
 
     my $attributes_values = $self->config->{'request_attribute_to_values'};
     my $code = $args->{service_code_alloy};
-    if (grep { $code =~ /$_/ } keys %{$self->config->{service_whitelist}->{'SC-Street Cleansing'}}) {
+    if (grep { $code =~ /$_/ } keys %{
```
dracos, line 36: I think this can be the slightly simpler:

```
... keys (
    %{$self->config->{service_whitelist}->{'SC-Street Cleansing'} || {} },
    %{$self->config->{service_whitelist}->{'Abandoned bikes/scooters'} || {} }
)
```
No need to put them in a hashref to take them back out? :)

MorayMySoc: I may be implementing your suggestion wrong but getting the error `Type of arg 1 to keys must be hash or array (not list)` when I don't make it a hashref

PR #398, opened 2025-02-24
[Bristol][Alloy] Quote grep to match category brackets

MorayMySoc:
The grep on category code should be an exact match and not a regexp

### DIFF
--- a/perllib/Open311/Endpoint/Integration/UK/Bristol/Alloy.pm
+++ b/perllib/Open311/Endpoint/Integration/UK/Bristol/Alloy.pm
@@ -33,7 +33,7 @@ sub process_attributes {
 
     my $attributes_values = $self->config->{'request_attribute_to_values'};
     my $code = $args->{service_code_alloy};
-    if (grep { $code =~ /$_/ } keys %{
+    if (grep { $code eq $_ } keys %{
             {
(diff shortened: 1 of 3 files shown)


perllib/Open311/Endpoint/Integration/UK/Bristol/Alloy.pm
```diff
@@ -33,7 +33,7 @@ sub process_attributes {
 
     my $attributes_values = $self->config->{'request_attribute_to_values'};
     my $code = $args->{service_code_alloy};
-    if (grep { $code =~ /$_/ } keys %{
+    if (grep { $code =~ /\Q$_\E/ } keys %{
```
dracos, line 36: Is there a reason this isn't an `eq`? Are we ever not looking for an exact match?

dracos (changes requested):

perllib/Open311/Endpoint/Integration/UK/Bristol/Alloy.pm
MorayMySoc: No - should be an exact match. I think

MorayMySoc: Just going to rewrite the commit message...

dracos (approved):

PR #400, opened 2025-03-13
[Symology] Open update CSVs in binary mode

davea:
Otherwise lines with extended chars and all subsequent lines are silently ignored.

For FD-5126


dracos (approved):
Just checking those are /meant/ to be the Unicode replacement character in the test files, and not an actual character like a pound symbol? (Just wondered what the encoding of the actual files are, I mean, if it's Unicode or Latin-1 or something.)

davea:
That's how the lines appear in the files we receive so figured it made sense to copy-paste one of the troublesome lines into our test files.

PR #401, opened 2025-03-14
Remove any use of my $x if $foo.

davea:
Counterpart to https://github.com/mysociety/fixmystreet/pull/5418


dracos (approved):
Ooh, thanks

PR #409, opened 2025-04-22
[Merton][Echo] Amend Bulky collection on same date

MorayMySoc:
When amending a bulky booking but keeping the same collection date, don't cancel the report.

Receive data from FMS and update the Event with
the changed details.

This isn't quite complete as not clear on the fields that need to be set for the Event and how they are updated rather than overwritten...hopefully the mechanics are ok though.

https://github.com/mysociety/societyworks/issues/4732

### DIFF
--- a/perllib/Open311/Endpoint/Integration/Echo.pm
+++ b/perllib/Open311/Endpoint/Integration/Echo.pm
@@ -538,4 +538,27 @@ sub update_event_payment {
     $args->{description} = ''; # Blank out so nothing sent to Echo now
 }
 
+=head2
+
+When a bulky booking has been updated, but the booking day hasn't changed,
+we want to update the current booking in place.
+
+=cut
+
+sub update_event_bulky_booking {
+    my ($self, $args) = @_;
+
+    my $integ = $self->get_integration;
+    my $event = $integ->GetEvent($args->{service_request_id}, $args->{id_type});
+    my $data = [ ];
+    foreach (@$fields) {
+    push @$data,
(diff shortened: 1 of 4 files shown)


perllib/Open311/Endpoint/Integration/Echo.pm
```diff
@@ -538,4 +538,27 @@ sub update_event_payment {
...
 
+=head2
+
+When a bulky booking has been updated, but the booking day hasn't changed,
+we want to update the current booking in place.
+
+=cut
+
+sub update_event_bulky_booking {
+    my ($self, $args) = @_;
```
dracos, line 549: So here (just documenting as much as anything), we will have, for Merton:
$args->{attributes}{ Bulky_Collection_Bulky_Items Bulky_Collection_Notes Exact_Location }
and for K/S:
$args->{attributes}{ TEM_-_Bulky_Collection_Description TEM_-_Bulky_Collection_Item Exact_Location }
and potentially images.

It feels like it would be easier for the update to wipe the existing items/images and replace them, rather than try and work out which ones have changed. So we need to construct a data field that empties out existing entries and then adds in new ones. I assume this is possible.

You can see in `post_service_request` that it loops through the extra data fields of the Echo event type (having fetched those with `GetEventType`) and for each one calls `check_for_data_value` to see if any supplied data matches that field. We could do the same here to 'fill in' the correct entries, or we could just for now hard code them for the different cobrands and improve it later one (happy to do that given matches the payment update above anyway).

PR #410, opened 2025-04-22
[Bristol] Fetch extra closure attribute info.

dracos:
Fetch Task Outcome and Removal Complete, and use those to construct an external status code.
Fo FD-5239


MorayMySoc (approved):
Looks good to me.

PR #411, opened 2025-04-28
[Aberdeenshire] Initial Confirm setup

davea:
For https://github.com/mysociety/societyworks/issues/4892

### DIFF
--- a/perllib/Open311/Endpoint/Integration/Confirm.pm
+++ b/perllib/Open311/Endpoint/Integration/Confirm.pm
@@ -70,14 +70,14 @@ has service_whitelist => (
 =head2 handle_jobs
 
 Whether cobrand fetches jobs from Confirm alongside enquiries. This is
-based on whether the cobrand has provided a GraphQL URL in its config.
+based on whether the cobrand has provided a list of services for jobs in its config.
 
 =cut
 
 has handle_jobs => (
     is => 'lazy',
     default => sub {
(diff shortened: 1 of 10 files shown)


perllib/Open311/Endpoint/Integration/Confirm.pm
```diff
@@ -806,9 +806,17 @@ sub get_service_requests {
...
+    # service/subject code - these are indicated with a _1/_2/etc suffix in
+    # service_whitelist. When fetching Enquiries from Confirm we join the
+    # service & subject code with _ and lookup the corresponding Open311
+    # service, which might fail if only the _1 suffixed version exists. So here
+    # we build a list of services to match against by stripping any _1/etc
+    # suffixes. As a result the first matching service will be used.
     my %services = map {
-        $_->{service_code} => $_
-    } @services;
+        (my $code = $_->{service_code}) =~ s/^([^_]*_[^_]*)_.*$/$1/;
```
dracos, line 817: Whilst it does look like cute smiley eyes, the other two places in this file that construct the suffix-less code both do so by split and join, so might be nicer to be consistent with those (could even factor it out to a function, it doesn't look like any of the intermediary info is used at all in any of the cases).

dracos (changes requested):
Only the one suggestion for consistency

dracos (approved):
ta!

PR #414, opened 2025-05-13
[Glos City] Additional questions to send to Alloy

nephila-nacrea:
For https://github.com/mysociety/societyworks/issues/4908.

Most additional questions take a Yes/No answer and we just append this to the report description.
The 'Dead animal' category, however, has answers that map to a specific Alloy attribute and we need to send through the correct code for the given answer.

### DIFF
--- a/perllib/Open311/Endpoint/Integration/AlloyV2.pm
+++ b/perllib/Open311/Endpoint/Integration/AlloyV2.pm
@@ -154,10 +154,30 @@ sub services {
         for my $subcategory (sort keys %{ $group_config }) {
             next if $subcategory eq 'alias';
             my $subcategory_config = $self->service_whitelist->{$group}->{$subcategory};
+
             my $subcategory_alias;
-            if (ref($subcategory_config) eq 'HASH' && exists($subcategory_config->{alias})) {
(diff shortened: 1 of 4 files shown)


chrismytton (changes requested):
I wonder if it might be more future-proof to look up the list of animal types in the Alloy API, rather than having them hard-coded in config? Basically the equivalent of this command in code:

```
bin/alloy/inspect-designs gloucester_alloy --list designs_animalTypeList_62fa576629e504015ad20fe1
```

Then you'd just need `designs_animalTypeList_62fa576629e504015ad20fe1 ` in the config. That way if they want to add/remove/update the list in the future it will automatically update FMS the next time the category list is refreshed rather than us needing to update the config manually.

(Aside: this was slightly awkward to review as the branch needs rebasing onto master, but I based my review on https://github.com/mysociety/open311-adapter/compare/mysociety:366437f...mysociety:203dfdf)

chrismytton (approved):
Discussed this in Slack with @nephila-nacrea, agreed it was ok to merge this as-is without my suggested changes, then can look at the changes as a potential future enhancement.

PR #415, opened 2025-05-14
Add config listing pages.

dracos:
This adds a page listing the council configs and letting you view them. The whole service is internal only, so this would be viewable if on the VPN.

### DIFF
--- a/perllib/Open311/Endpoint.pm
+++ b/perllib/Open311/Endpoint.pm
@@ -10,7 +10,9 @@ use Encode qw(encode_utf8);
 use Web::Simple;
 
 use JSON::MaybeXS;
+use Path::Tiny;
 use XML::Simple;
+use YAML::XS qw(LoadFile Dump);
 
 use Open311::Endpoint::Result;
 use Open311::Endpoint::Service;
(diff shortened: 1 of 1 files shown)


perllib/Open311/Endpoint.pm
```diff
@@ -340,6 +342,16 @@ sub dispatch_request {
         my ($self, $service_request_id, $args) = @_;
         $self->call_api( GET_Service_Request => $service_request_id, $args );
     },
+
+    sub (GET + /config/) {
```
davea, line 346: Just because I know I will get this wrong 100% of the time - any way the trailing slash could be removed/optional?

perllib/Open311/Endpoint.pm
```diff
@@ -733,6 +745,68 @@ sub GET_Service_Request {
...
+    %$conf = map {
+        if (/secret|password|(?<!data_)key|token/) {
+            $_ => '***redacted***';
+        } else {
+            $_ => $conf->{$_};
+        }
+    } keys %$conf;
+
+    return [
+        200, [ 'Content-Type' => 'application/json' ], [ $self->json->canonical->encode($conf) ],
```
davea, line 806: Could this output YAML rather than JSON? Or is JSON a deliberate choice? I know it's a subset of YAML but I find the whitespace-delimited version much easier on the eye.

perllib/Open311/Endpoint.pm
```diff
@@ -733,6 +739,66 @@ sub GET_Service_Request {
...
+sub GET_config_entry {
+    my ($self, $jurisdiction) = @_;
+
+    my $path = path(__FILE__)->parent(3)->realpath->child('conf');
+    $path = $path->child('council-' . $jurisdiction. '.yml');
+    return "Config not found" if !$path->is_file;
+
+    my $conf = LoadFile($path);
+    %$conf = map {
+        $_ => (/secret|password|(?<!data_)key|token/) ? '***' : $conf->{$_}
```
davea, line 796: Ooh sorry just spotted that Rutland uses `credentials`, maybe that should be captured here too.

dracos: ah yes, only in staging, that's why I missed it :+1:

PR #418, opened 2025-05-16
[Confirm] More thorough check for handle_jobs

davea:


dracos (commented):
Not sure this makes any difference, as e.g. shropshire had web/tenant for photo upload, but doesn't have jobs (if I understand it right?) so would have to have taken out job_service_whitelist anyway as I did?

PR #419, opened 2025-06-03
[Gloucester] Add external_status_code to updates

chrismytton:
Switch the test over to use inspections instead of defects, disable defects, and override `_get_inspection_status` to fetch the status from Alloy for use in the `external_status_code` field on updates.

Fixes https://github.com/mysociety/societyworks/issues/4972

## Notes to deployer

Complete the following when deploying:

- [ ] Update the Alloy configuration to use inspections instead of defects (should be similar/same as the diff for the test yml in this PR.


nephila-nacrea (approved):
Looks good to me, pretty much follows the pattern for other _get_inspection_status() implementations 👍

PR #421, opened 2025-06-11
[Ezytreev] Fix 'GetEnquiryChanges' calls.

neprune:
The API started requiring a crmXRef on these requests around early May 2025.

https://mysocietysupport.freshdesk.com/a/tickets/5580


dracos:
To note, this doesn't/wouldn't work because `get_service_request_updates` isn't passed a `service_request_id_ext` (it's called in `perllib/Open311/Endpoint/Role//mySociety.pm`), only a time window to look up updates in. I see this has been superseded by a time-based API call anyway, but just for info.

PR #424, opened 2025-06-25
[Gloucester] Send through priority and target date 

dracos:
Bit of duplication of FMS working day code, but seemed easier to hold it all within open311-adapter.

Server config would be as in tests, but for all the categories that needed to be High and with all the target SLAs.

Fixes https://github.com/mysociety/societyworks/issues/5018


chrismytton (approved):
Might be good to extract the `WorkingDays` module into commonlib, or even CPAN (if that's straightforward) at some point, but I think this is fine for now.

PR #425, opened 2025-07-01
Bucks cams integration

MorayMySoc:
Integration between FMS and CAMS Desktop for Buckinghamshire as per the tech specification.

https://docs.google.com/document/d/1X3v_y2TBnbaTqDwMlo-sUwt_8RganhJHge-OXleZS-8/edit?tab=t.0#heading=h.a88nmmw52k7z

https://github.com/mysociety/societyworks/issues/4979
https://github.com/mysociety/societyworks/issues/4980
https://github.com/mysociety/societyworks/issues/5014
https://github.com/mysociety/societyworks/issues/5015
https://github.com/mysociety/societyworks/issues/5013

### DIFF
--- a/perllib/Integrations/Rest.pm
+++ b/perllib/Integrations/Rest.pm
@@ -1,6 +1,7 @@
 package Integrations::Rest;
 
 use Moo;
+use HTTP::Request::Common;
 use LWP::UserAgent;
 use JSON::MaybeXS;
 
@@ -47,7 +48,12 @@ has json => (
 
 api calls are either GET or POSTING JSON data.
 
-Expects { call => ** string uri **, $body => ** hashref of data ** headers => ** hashref of headers ** }
+Expects {
+    call => ** string uri **,
+    headers => ** hashref of headers **
+    body => ** hashref of data to be JSON-encoded, or **
+    form => ** hashref or arrayref of form data **
+}
 
 JSON is expected in a successful response, but allow_nonref can be set in initialisation.
 
@@ -58,7 +64,9 @@ sub api_call {
 
     my $call = $args{call};
     my $body = $args{body};
+    my $form = $args{form};
     my $headers = $args{headers};
+
     $self->logger->debug($call);
 
     my $ua = LWP::UserAgent->new(
@@ -67,7 +75,9 @@ sub api_call {
     );
 
     my $method = $args{method};
-    $method = $body ? 'POST' : 'GET' unless $method;
+    $method = ($body || $form) ? 'POST' : 'GET' unless $method;
+    $method = HTTP::Request::Common->can($method);
+
     my $uri = URI->new( $self->config->{api_url} . $call );
     $uri->query_form(%{ $args{params} });
 
@@ -86,6 +96,7 @@ sub api_call {
         }
     }
 
+    my $request = $method->($uri, $form ? ($form) : (), %$headers);
     my $response = $ua->request($request);
     if ($response->is_success) {
         $self->logger->debug($response->content);
--- a/perllib/Integrations/Rest.pm
+++ b/perllib/Integrations/Rest.pm
@@ -1,0 +1,108 @@
+package Integrations::Rest;
+
+use Moo;
+use LWP::UserAgent;
+use JSON::MaybeXS;
+
+with 'Role::Config';
+with 'Role::Logger';
+
+=head2 caller
+
+This is a generic JSON integrater and needs to register which integration it is being called
+for to identify in error logs
+
+=cut
+
+has caller => (
+    is => 'ro',
+    required => 1,
+);
+
+=head2 allow_nonref
+
+Set whether json decoding allows non json data (ie a string)
+to be parsed - defaults to 0
+
+=cut
+
+has allow_nonref => (
+    is => 'ro',
+    default => 0
+);
+
+=head2 json
+
+JSON class to use which can be set by integration, but will default to
+MaybeXS and if so, is set to use the attribute allow_nonref setting
+
+=cut
+
+has json => (
+    is => 'lazy',
+    default => sub { JSON->new->utf8->allow_nonref($_[0]->allow_nonref) },
+);
+
+=head2 api_call
+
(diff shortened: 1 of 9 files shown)


t/open311/endpoint/buckinghamshire_cams.yml
```diff
@@ -0,0 +1,35 @@
...
+
+service_list:
+  Public Rights of Way:
+    '9':
+      service_name:
+        - 'Alignment/Path Off Line'
+    'I274':
+      service_name:
+        - 'Bridge/Damaged'
+        - 'Damaged Bridge'
```
chrismytton, line 20: I wonder if it might be clearer having these two values under separate keys, rather than having the 0th element be the CAMS category and the 1st element be the FMS category. Could still have the fallback to the CAMS name if there is no FMS category.

Having separate keys would also make this more self-documenting, rather than having to look at the code to understand the meaning of the element order.

perllib/Open311/Endpoint/Integration/Cams.pm
```diff
@@ -0,0 +1,422 @@
...
+
+    my $TypeDescr;
+    for my $category (keys %{ $self->service_list }) {
+        for my $id (keys %{ $self->service_list->{$category}}) {
+            if ($id eq $args->{service_code}) {
+                $TypeDescr = $self->service_list->{$category}->{$id}->{service_name}[0];
+                last;
+            }
+        }
+    }
```
chrismytton, line 250: Might be worth having a check after this loop to ensure `$TypeDescr` has a value and if not then `die` with a useful message so that it becomes a stuck report.

perllib/Open311/Endpoint/Integration/Cams.pm
```diff
@@ -0,0 +1,422 @@
...
+        }
+
+        @$recent_updates = grep {
+            $_->{'LastUpdatedDate'} >= $start_time
+            && $_->{'LastUpdatedDate'} <= $end_time
+        } @$recent_updates;
+
+        my $ug; my $uuid;
+        for my $update (@$recent_updates) {
+            $ug = Data::UUID->new;
```
chrismytton, line 331: I think this could be moved outside the loop, I don't believe it maintains any state between calls to `create()`.

chrismytton (changes requested):
Lots of good stuff in here, definitely heading in the right direction!

Left a few minor comments, but I think the main point for me is having the FMS category and the CAMS category in a single list under `service_name` is a bit confusing.

t/open311/endpoint/buckinghamshire_cams.yml
MorayMySoc: cf72785c614dfe627789361d4bbc7b6a80bf36b9

perllib/Integrations/Rest.pm
```diff
@@ -0,0 +1,100 @@
...
+        agent => "FixMyStreet/open311-adapter",
+        timeout => 5*60,
+    );
+
+    my $method = $args{method};
+    $method = $body ? 'POST' : 'GET' unless $method;
+    my $uri = URI->new( $self->config->{api_url} . $call );
+    $uri->query_form(%{ $args{params} });
+
+    my $request = HTTP::Request->new($method, $uri);
```
dracos, line 64: We probably want to use HTTP::Request::Common here so it handles multipart uploads etc for us when necessary

MorayMySoc:
@dracos 649dadd9a288e0ed55983d2fd5f2de8d53c1f69d is to convert to using an md5 hash for the id to remove ids coming through with unique ids multiple times as we  discussed. I think this confirms that the updated date field is ok for this: https://3.basecamp.com/4020879/buckets/42055227/todos/9045756517

perllib/Integrations/Rest.pm
```diff
@@ -0,0 +1,107 @@
...
+                $headers->{Content_Type} = 'form-data';
+            }
+        }
+    }
+
+    my $request = $method->($uri, $form ? ($form) : (), %$headers);
+    my $response = $ua->request($request);
+    if ($response->is_success) {
+        $self->logger->debug($response->content);
+        return decode_json($response->content, $self->allow_nonref);
```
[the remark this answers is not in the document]
MorayMySoc: 8218985a5653c201ae386124cbda0e6074f1228c

perllib/Integrations/Rest.pm
```diff
@@ -88,13 +100,13 @@ sub api_call {
     my $response = $ua->request($request);
     if ($response->is_success) {
         $self->logger->debug($response->content);
-        return decode_json($response->content, $self->allow_nonref);
+        return $self->json->decode($response->content, $self->allow_nonref);
```
dracos, line 103: You don't want the second argument here now

perllib/Integrations/Rest.pm
```diff
@@ -88,13 +100,13 @@ sub api_call {
...
     } else {
         $self->logger->error($call);
-        $self->logger->error(encode_json($body)) if $body and (ref $body eq 'HASH' || ref $body eq 'ARRAY');
+        $self->logger->error($self->json->encode($body)) if $body and (ref $body eq 'HASH' || ref $body eq 'ARRAY');
         $self->logger->error($response->content);
         try {
-            my $json_response = decode_json($response->content, $self->allow_nonref);
+            my $json_response = $self->json->decode($response->content, $self->allow_nonref);
```
dracos, line 109: Or here

perllib/Open311/Endpoint/Integration/Cams.pm
```diff
@@ -300,8 +300,7 @@ sub post_service_request {
         body => $serviceRequest,
         headers => { '.aspxauth' => $self->access_token }
     );
-
-    if ($response && $response ne 'fail') {
+    if ($response && $response =~ /\d+/) {
```
dracos, line 303: `/^\d+$/`? In case the response is an HTML page with "404 not found" on it or whatever they break

dracos (changes requested):

PR #426, opened 2025-07-01
[Bexley] Bulky collection submission

dracos:
For https://github.com/mysociety/societyworks/issues/5021
For https://github.com/mysociety/societyworks/issues/5022

PR #433, opened 2025-08-05
[Bexley][Bulky] Custom worksheet message

nephila-nacrea:
For https://3.basecamp.com/4020879/buckets/42017522/card_tables/cards/8904241999 .

### DIFF
--- a/perllib/Open311/Endpoint/Integration/UK/Bexley/Whitespace.pm
+++ b/perllib/Open311/Endpoint/Integration/UK/Bexley/Whitespace.pm
@@ -12,16 +12,42 @@ around BUILDARGS => sub {
 sub _worksheet_message {
     my ($self, $args) = @_;
 
-    my @messages;
-    foreach (
+    my @attributes = $args->{service_code} eq 'bulky_collection' ? (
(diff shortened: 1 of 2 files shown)


perllib/Open311/Endpoint/Integration/UK/Bexley/Whitespace.pm
```diff
@@ -12,16 +12,42 @@ around BUILDARGS => sub {
...
+            );
+            my $dtf_to = DateTime::Format::Strptime->new(
+                pattern => '%d/%m/%Y',
+            );
+
+            # Undef if error
+            $val = $dtf_from->parse_datetime($val);
+            $val = $dtf_to->format_datetime($val) if $val;
+
+            unless ($val) {
```
dracos, line 44: I think if we get here with a bad date, we should just go ahead and submit with whatever we originally got rather than error, they'll presumably be able to understand it should this happen and that's better than it not getting sent.

dracos (changes requested):
One tiny change is all

nephila-nacrea:
Mucked up the merge/rebases a bit, this has been merged into https://github.com/mysociety/open311-adapter/pull/426 .

PR #438, opened 2025-09-05
[Brent][Atak] Update for sending to MSS

MorayMySoc:
Removes all fetching of updates.

Updates field names for posting data.

Updates tests.

Have left the Atak file naming as probably want
categories to carry on as before.

https://github.com/mysociety/societyworks/issues/5027


chrismytton (approved):
Looks good! I think this would have been a bit easier to review if it was split into two commits, one for removing the updates stuff and one for changing the new report bits, don't worry about it now, but something for next time 🙂

PR #441, opened 2025-09-11
[Confirm] Support 'supersedes' field for defects and their updates.

neprune:
Open311 component of the work for https://github.com/mysociety/societyworks/issues/5138.


davea (approved):
Looks good! I removed the config option for `include_supersedes_field_for_defects` as I couldn't think of a reason why we shouldn't always query for it, and just populate `extras` if there was a value. Hope that's OK!

PR #447, opened 2025-09-29
[Hampshire][Confirm] Adds an allowed_attributes attribute

MorayMySoc:
When this exists in config, it works to only allow attributes included in the list. It is used in place of ignore_attributes which will be discounted if it also exists.

If there is an empty list in allow_attributes, it will mean that no attributes will be passed through.

https://github.com/mysociety/societyworks/issues/5102


perllib/Open311/Endpoint/Integration/Confirm.pm
```diff
@@ -234,6 +234,24 @@ has ignored_attributes => (
...
+through of attributes to only pull through those in this list and
+will act instead of ignored_attributes.
+
+An empty array will, in effect, act as though all attributes
+had been added to ignored_attributes.
+
+=cut
+
+has allowed_attributes => (
+    is => 'ro',
```
dracos, line 252: If this additionally had something the same as`jurisdiction_ids` does in `Endpoint.pm` (without the default line), (and `use MooX::HandlesVia;` and `use Types::Standard qw(Maybe ArrayRef);`), then instead of `$self->allowed_attributes && ref $self->allowed_attributes eq 'ARRAY'` you could have `$self->has_allowed_attributes` and instead of `@{$self->allowed_attributes}` you could have `$self->get_allowed_attributes`.

dracos (changes requested):
Looks good, one suggestion to use the new attribute more

dracos (approved):
Nice :)

PR #477, opened 2026-02-20
[Bartec] Add ability to update status and extra data.

dracos:
For https://github.com/mysociety/societyworks/issues/5312

Adds a forward_status_map to cover state changes from FMS to Bartec (and creating new report uses that rather than an inverted status_map (which is used for the other way)) using status set, and updates the service request with any extra data provided.


chrismytton (approved):
Minor comment on an edge-case, but other than that looks good!

chrismytton (approved):

PR #487, opened 2026-03-09
[Dumfries] Alloy defect fetching & photo handling improvements

davea:
This is the second of three PRs for the Dumfries Alloy integration. This PR:

 - Makes improvements to report sending & update fetching
 - enables Defect fetching
 - adds inspection & job photo fetching

A lot of this code gets refactored/improved in the next PR. For context, the overall diff of all three PRs to `master` is here: https://github.com/mysociety/open311-adapter/compare/master...dumfries-alloy-pr-3?expand=1

### DIFF
--- a/perllib/Open311/Endpoint/Integration/AlloyV2.pm
+++ b/perllib/Open311/Endpoint/Integration/AlloyV2.pm
@@ -266,8 +266,7 @@ sub post_service_request {
     # Upload any photos to Alloy and link them to the new resource
     # via the appropriate attribute
     if (@$files) {
-        my $attributes = [{ attributeCode => $self->config->{resource_attachment_attribute_id}, value => $files }];
-        $self->_update_item($item_id, $attributes);
+        $self->_attach_files_to_service_request($item_id, $files);
     }
 
     # create a new Request and return it
@@ -275,6 +274,13 @@ sub post_service_request {
         service_request_id => $item_id
     );
 
+}
+
+sub _attach_files_to_service_request {
+    my ($self, $item_id, $files) = @_;
+
+    my $attributes = [{ attributeCode => $self->config->{resource_attachment_attribute_id}, value => $files }];
+    $self->_update_item($item_id, $attributes);
 }
 
 =head2 _get_service_code
@@ -740,7 +746,8 @@ sub _get_inspection_updates {
     }
     my @updates;
     foreach (@$resources) {
-        push @updates, $self->_get_inspection_updates_design($_, $args);
+        my ($design_updates, $items_by_id) = $self->_get_inspection_updates_design($_, $args);
+        push @updates, @$design_updates;
     }
     return @updates;
 }
@@ -754,7 +761,7 @@ sub _get_inspection_updates_design {
     my @updates;
 
     my $mapping = $self->config->{inspection_attribute_mapping};
-    return () unless $mapping;
+    return ([], {}) unless $mapping;
 
     my %join;
     if (my $extra_mapping = $mapping->{extra_attributes}) {
@@ -772,6 +779,7 @@ sub _get_inspection_updates_design {
 
     my $assigned_to_users = $self->get_assigned_to_users(@$updates);
 
+    my %items_by_id;
     for my $report (@$updates) {
         next unless $self->_accept_updated_resource($report, $design);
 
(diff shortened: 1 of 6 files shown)


perllib/Open311/Endpoint/Integration/UK/Dumfries.pm
```diff
@@ -128,6 +200,132 @@ sub _skip_inspection_update {
...
+                $report, 'attributes_defectsWithInspectionsDefectInspection', $args, $cache
+            );
+
+    # For each update, fetch the associated resource and add media URLs
+    # Also handle special case for latest_inspection_time
+    for my $update (@updates) {
+        my $service_request_id = $update->service_request_id;
+
+        # Fetch the resource to get job and inspection media URLs
+        my $response = $self->alloy->api_call(call => "item/$service_request_id");
```
dracos: Noting my comment at https://github.com/mysociety/societyworks/issues/5443#issuecomment-3998262553 : "[Aside note: we have had issues in the past with the number of Alloy API calls, e.g. https://github.com/mysociety/societyworks/issues/3710 and now I've looked at this I'm a bit worried about that. Certainly for example, as we've already fetched all the items in the original get_inspection_updates_design call, we shouldn't be calling item/* for each of them again in another loop. And can we fetch the joined jobs/inspections in the original AQS query for updated resources, not 2N calls to look them up after?"

perllib/Open311/Endpoint/Integration/UK/Dumfries.pm
```diff
@@ -58,6 +62,127 @@ sub _get_service_code {
...
+
+    my $attributes = $self->alloy->attributes_to_hash($defect);
+    my $service_code = $attributes->{$mapping->{service_code}};
+    
+    $service_code = $service_code->[0] if ref $service_code eq 'ARRAY';
+    
+    return $service_code;
+}
+
+=head2 service
```
dracos: This doesn't seem linked to this commit which is about defect fetching?

perllib/Open311/Endpoint/Integration/UK/Dumfries.pm
```diff
@@ -481,36 +572,120 @@ sub _find_latest_inspection {
...
+    } catch {
+        $self->logger->error("Failed to fetch photo $item_id: $_");
+        return [ 404, [ 'Content-Type', 'text/plain' ], [ 'Photo not found' ] ];
+    };
+
+    return [ 404, [ 'Content-Type', 'text/plain' ], [ 'Photo not found' ] ] unless $content;
+    return [ 200, [ 'Content-Type', $content_type ], [ $content ] ];
+}
+
+=head2 _post_creation_processing
```
dracos: This function doesn't exist/isn't called until 0eba07f5.

dracos (changes requested):
The code changes overall look generally fine and I assume are doing as is desired - one or two small questions is all.

In meta terms, for looking back at this in future, there are a number of issues with functions not existing when things that use them are, and other things like that, presumably artifacts introduced during a rebase or the like. I don't really want to say squash it all together, as it's quite a lot of code and I think there is value in some separate commits for the different functionality, but it's going to need some tidying up to e.g. remove code being added and then deleted, or name changed multiple times back and forth e.g. _get_inspection_status adds a second return and breaks in 4408ace7 then the function inside changes name 4 times in further commits). Perhaps bringing all the job completion photo stuff together in one commit would help sort that out. If this can't be done before going live, I think perhaps we should run open311-adapter on a branch including this reviewed code until that can be done as soon as possible after.

dracos (changes requested):
Thanks, that is much cleaner, isn't it :)

One line that needs using later on, I think; then only other thing is the `_post_creation_processing` function in 53c348f02c51b24db07bfbd93e4aafe973f4ddf9 shouldn't be in that commit (which doesn't mention it) but should be in the first commit of pr-3.

perllib/Open311/Endpoint/Integration/AlloyV2.pm
```diff
@@ -1028,6 +1038,8 @@ sub _get_service_requests_resource {
         my %args;
 
         next if $self->skip_fetch_defect( $request );
+        my $attributes = $self->alloy->attributes_to_hash($request);
+        my $title = $attributes->{attributes_itemsTitle} || 'Unknown title';
```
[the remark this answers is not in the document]
davea: Was needed by a later commit (36bd0c68ccb0edf9da9cfa97ac5d670e6f014f23), so moved it there.

dracos (approved):

davea:
Merged as part of https://github.com/mysociety/open311-adapter/pull/459

PR #499, opened 2026-03-27
[Rutland][SalesForce] Rutland salesforce refresh

MorayMySoc:
Adds category whitelisting for Rutland Salesforce to filter out duplicates when Confirm integration goes live

Remove 'hints' feature so that extra data is now passed to be a notice in FMS.

https://github.com/mysociety/societyworks/issues/5400


dracos (approved):
Looks good; wonder if we know the list from Confirm it'd be better as a list of services to ignore, rather than include?

MorayMySoc:
The category names aren't exactly the same I don't think so we've got the SF ones to keep in the spec

PR #502, opened 2026-04-08
[Rutland][Confirm] Fix minor bug and change match

MorayMySoc:
Two commits - one to fix an issue where Confirm doesn't order its updates by date, so we will now do that.
https://github.com/mysociety/societyworks/issues/5494

The other just adds DT10 to photo tags to be picked, so it is both DT10 and DT20
https://github.com/mysociety/societyworks/issues/5416


t/open311/endpoint/rutland_confirm.t
```diff
@@ -42,7 +42,7 @@ subtest "Only uses the photo with the correct classification tag" => sub {
             URL => '1',
             Name => '1.jpg',
             Date => DateTime->now->subtract(days => 1),
-            ClassificationCode => 'DT10',
+            ClassificationCode => 'DT30',
```
chrismytton, line 45: I think this means that the DT10 acceptance change isn't actually being tested. Might be worth adding a DT10 photo to the fixture and bumping the count to 2, so the test would catch a regression back to eq 'DT20'.

Something like:

```diff
             Name => '3.jpg',
             Date => DateTime->now->subtract(days => 3),
         },
+        {
+            URL => '4',
+            Name => '4.jpg',
+            Date => DateTime->now->subtract(days => 4),
+            ClassificationCode => 'DT10',
+        },
     );
     my @filtered = $endpoint->filter_photos_graphql(@photos);
 
-    is @filtered, 1;
-    is $filtered[0]->{URL}, 2;
+    is @filtered, 2;
+    my %by_url = map { $_->{URL} => 1 } @filtered;
+    ok $by_url{2}, 'DT20 photo included';
+    ok $by_url{4}, 'DT10 photo included';
 };
 
 subtest 'Only pass on log notes for updates for "FMS" status' => sub {
```

chrismytton (approved):
Look good, just one minor comment on the tests.

PR #503, opened 2026-04-09
[Aurora] Handle photo uploads correctly

davea:
Changed test to look at file size as temp path is unpredictable.

For https://github.com/mysociety/societyworks/issues/5514


neprune (approved):
nice

PR #504, opened 2026-04-10
[Bartec] Reduce ServiceRequests_History_Get API calls

davea:
We were making API calls to fetch history for service requests from other CRMs and for service request types that we don't handle.

For FD-6864

### DIFF
--- a/perllib/Open311/Endpoint/Integration/Bartec.pm
+++ b/perllib/Open311/Endpoint/Integration/Bartec.pm
@@ -102,7 +102,7 @@ sub _get_service_name {
     my ($self, $service) = @_;
 
     $service->{Description} =~ s/(.)(.*)/\U$1\L$2/;
-    (my $class = $service->{ServiceClass}->{Description}) =~ s/(.)(.*)/\U$1\L$2/;
+    (my $class = $service->{ServiceClass}->{Name}) =~ s/(.)(.*)/\U$1\L$2/;
     my $service_name = $service->{Description};
 
     # service type names are not unique in bartec so need to distinguish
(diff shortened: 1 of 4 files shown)


perllib/Open311/Endpoint/Integration/Bartec.pm
```diff
@@ -439,6 +439,18 @@ sub get_service_request_updates {
...
     my $updates = $self->get_integration->_coerce_to_array( $response, 'ServiceRequest_Updates' );
     for my $update ( @$updates ) {
+        my $ref = $update->{JobReference} // ''; # JobReference is the external (to Bartec) ID, i.e. FMS report ID
+        next unless $ref =~ /^\d{7,}$/; # skip ServiceRequest if its job ref doesn't look like an FMS ID
+
+        # Skip updates for ServiceRequests whose services we're not responsible for
+        my $service = {
+            Description => $update->{ServiceType},
+            ServiceClass => {
+                Description => $update->{ServiceClass},
```
dracos: This won't work, I'm afraid, because the $update-{ServiceClass} here is the name (e.g. FLY TIP) and the description (which is what is used in the config) is "FLT TIP" (see e.g. response to ServiceRequests_Types_Get with Class 275 Name GENERAL, Description GENERAL but service class 86, Name FLY TIP, Description FLT TIP). I don't know whether it's possible to change `services` (I think that's only use?) to use the name rather than the description, if that is the only one that 'differs' like that, it would probably make sense and then this would work.

dracos (changes requested):
An issue with the name/description difference that will need resolving.

I believe we are looking back 2 hours, perhaps they have a timezone issue..., so I guess we could give them their own bin/fetch in the cron? Like we already do for Lincs (look back 15 minutes) and Nland (look back 30m every 10m). Could look back 0.5hr (so if they do have timezone bug that'd be 1.5hr) but still every 5 minutes?

davea:
Thanks, good suggestion. There were a few others that needed changing in `service_map` in config, but it does seem to work as intended:

```
Offensive graffiti -> Off. graffiti
Non offensive graffiti -> Non-off. graffiti
Flt tip -> Fly tip
Food bag request -> Food bags
Bin not returned to collection point -> Not returned to cp
Additonal bin requests -> Additional bin
```

dracos (approved):
Woot

PR #505, opened 2026-04-14
[Lincolnshire] Use existing photo_filter method for photos from GraphQL

davea:
For FD-6873

### DIFF
--- a/perllib/Open311/Endpoint/Integration/Confirm.pm
+++ b/perllib/Open311/Endpoint/Integration/Confirm.pm
@@ -632,7 +632,7 @@ sub get_service_request_updates {
         my $job_photos = @{ $integ->enquiry_update_job_photo_statuses };
         my $defect_photos = @{ $integ->enquiry_update_defect_photo_statuses };
 
-        my $documents = 'documents { url documentName documentDate docTypeCode}';
+        my $documents = 'documents { url documentName documentDate docTypeCode documentNotes}';
         my $job_documents = $job_photos ? $documents : "";
         my $defect_documents = $defect_photos ? $documents : "";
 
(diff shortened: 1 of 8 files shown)


perllib/Open311/Endpoint/Integration/Confirm.pm
```diff
@@ -941,14 +941,29 @@ sub construct_photo_url_from_graphql_fetched_doc {
...
+    Notes => 'document notes'
+}
+
+Subclasses e.g. Lincolnshire can make further determinations based on the Notes field.
+=cut
+
 sub photo_filter {
     my ($self, $doc) = @_;
-    return $doc->{fileName} =~ /jpe?g/i;
+    return $doc->{Name} =~ /\.(jpg|jpeg|pjpeg|gif|tiff|png)$/i;
```
davea, line 961: I don't recall why we only allowed JPEGs here when using the web API but a broader range when using GraphQL.

dracos (changes requested):
Looks good - for completeness, could we update the Rutland function to use photo_filter instead, just in case it had to be switched to the connector for some reason?

I assume Aberdeenshire is bit harder to change given what it does, but a cross-solution there would be good - actually, could the one direct call/grep to photo_filter (for connector) in job_photo_urls_for_enquiry_update be replaced by a call to filter_photos_graphql() (that could then be renamed!) and then Aberdeenshire's code would catch both ways as-is.

perllib/Open311/Endpoint/Integration/Confirm.pm
dracos: I assume this would be okay

dracos (approved):
Woo, should prevent this happening again :)

PR #506, opened 2026-04-15
[Central Beds] Do not pull in updates for Jadu

nephila-nacrea:
We do not want to pull in updates as we mark reports in Jadu category (i.e. Fly Tipping) as closed automatically; pulling in updates reopens them with the 'investigating' status.

### DIFF
--- a/perllib/Open311/Endpoint/Integration/UK/CentralBedfordshire/Jadu.pm
+++ b/perllib/Open311/Endpoint/Integration/UK/CentralBedfordshire/Jadu.pm
@@ -545,6 +545,12 @@ Reads updates from C<update_storage_file>.
 
 sub get_service_request_updates {
     my ($self, $args) = @_;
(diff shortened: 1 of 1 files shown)


neprune (changes requested):
Sorry to be a pain, but I thought actually maybe better to have this managed as a flag we manage in config. So can keep the tests as is. Also looks like some other changes got caught up  - just need a rebase maybe?

nephila-nacrea:
> Sorry to be a pain, but I thought actually maybe better to have this managed as a flag we manage in config. So can keep the tests as is. Also looks like some other changes got caught up - just need a rebase maybe?

Ugh, I had rebased on top of staging then pushed.
Have made it a config change; added the flag in mysociety-servers and deployed everything to staging, to make sure it all works.

neprune (approved):

PR #507, opened 2026-04-20
[Alloy] Have an ID key way of having categories.

dracos:
Switch Dumfries / Gloucester / Hackney to use it.

For https://github.com/mysociety/societyworks/issues/5333
And actually for https://mysocietysupport.freshdesk.com/a/tickets/6863

Hackney want two new subcategories (in same parent) to go to the same Alloy category ID.
This seemed quite challenging (doesn't work as is; can't use alias, as same Alloy name), so was trying to do some of https://github.com/mysociety/societyworks/issues/5333 but that led to issues with Gloucester as mentioned there, so ended up with a new way of specifying, which then does let us have dupe IDs going to a name. Hopefully makes sense!

(Would involve updating those three in server config to match new way, of course. I think this would be fine from FMS point of view, the names would all still match so nothing should change beside the destination codes.)


perllib/Open311/Endpoint/Integration/AlloyV2.pm
```diff
@@ -145,26 +150,24 @@ sub services {
...
             next if $subcategory eq 'alias';
             my $subcategory_config = $self->service_whitelist->{$group}->{$subcategory};
 
-            my $subcategory_alias;
             my @questions;
-
-            if ( ref($subcategory_config) eq 'HASH' ) {
-                if ( exists( $subcategory_config->{alias} ) ) {
-                    $subcategory_alias = $subcategory_config->{alias};
+            if ($subcategory =~ /^[0-9a-f]{24}$/) {
```
davea: Because this doesn't allow the `_1` suffix, you can end up with the raw service code in `service_name` in the `services.xml`/`services.json` output. The Gloucester tests don't cover it but if you add this test around L114 of `gloucester_alloy.t` you can see the problem:

```perl
subtest 'check service_name' => sub {
    my $res = $endpoint->run_test_request(
        GET => '/services.json' );
    ok $res->is_success, 'json success';

    my $services = decode_json( $res->content );

   my %services = map { $_->{service_code} => $_ } @$services;
   is $services{'61b9e1ccfb9e760158036bc1_3'}->{service_name}, 'Damaged litter bin';
};
```

davea (changes requested):

davea (approved):
Looks good!

PR #510, opened 2026-04-30
[Open311][Confirm] Document Confirm settings

MorayMySoc:
Guide to what's required for Confirm integration
settings

https://github.com/mysociety/societyworks/issues/5329

[skip changelog]

### DIFF
--- a/conf/council-rutland_confirm.yml-example
+++ b/conf/council-rutland_confirm.yml-example
@@ -1,18 +1,42 @@
+# General for integrations
+
 "logfile": ""
 "min_log_level": ""
+"cutoff_enquiry_date": ""
 
+# Confirm settings
+
+# For sending/fetching reports/updates
+# Required
 "endpoint_url": ""
 "username": ""
 "password": ""
 "tenant_id": ""
+# Required for handling photos
+"web_url": ""
+"uploads_dir": ""
 
-"cutoff_enquiry_date": ""
-"enquiry_method_code": ""
-"point_of_contact_code": ""
+# Usually Europe/London
+# Required
 "server_timezone": ""
 
+# If we're fetching jobs and defects
+# then required, but also
+# adds a speed boost for other
+# fetching of enquiries and photos
+"graphql_pass": ""
+
+# These may be optional depending on
+# Confirm setup but documented in
+# wiki how to fetch them if not
+# supplied
+"enquiry_method_code": ""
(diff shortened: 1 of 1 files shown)


conf/council-rutland_confirm.yml-example
```diff
@@ -1,18 +1,37 @@
...
+"graphql_pass": ""
+"web_url": ""
+"uploads_dir": ""
 
-"reverse_status_mapping": {}
+# These may be optional depending on
+# Confirm setup but documented in
+# wiki how to fetch them if not
+# supplied
+# Maybe required
```
dracos: Probably doesn't need this line as says may be optional just above

conf/council-rutland_confirm.yml-example
```diff
@@ -1,18 +1,37 @@
...
+# Usually Europe/London
+# Required
 "server_timezone": ""
 
-"service_whitelist": {}
-
-"forward_status_mapping": {}
+# For further features such as fetching photos with reports
+# need to add graphql and have set directory for storing photos
+# Required for extra features
```
dracos: I think this needs some expansion.

uploads_dir and web_url are needed in order to send photos in to Confirm, don't need GraphQL access for that, and we always require that as a core thing, assuming they want photos.

GraphQL is needed to fetch jobs and defects, not just enquiries (they have handle_jobs/handle_defects flags in the code, which depend upon job_service_whitelist and defect_service_mapping being set, looks like). And if graphql auth is present, it is also used to speed up fetching enquiry updates and completion photos as well. GraphQL being present also currently means category sync is on but I'm adding a feature flag for that in a different PR, because you might not want it to be.

conf/council-rutland_confirm.yml-example
```diff
@@ -1,18 +1,37 @@
+# General for integrations
+
 "logfile": ""
 "min_log_level": ""
+"cutoff_enquiry_date": ""
+"service_whitelist": {}
+"forward_status_mapping": {}
+"reverse_status_mapping": {}
+
+# Confirm settings
```
dracos, line 7: I think we generally put these first, before the service whitelist and status mappings, as they're short (unlike the main config).

dracos (changes requested):
Few suggestions :)

dracos (approved):
👍  Thanks

PR #511, opened 2026-05-08
Allow extra details to be updated without an assigned user

m7xedhoward:
Allow extra details to be updated without an assigned user


m7xedhoward:
#513

PR #512, opened 2026-05-12
[Dumfries] Setting 'responsibility' on the extras of an update

MorayMySoc:
If there is a responsibility attribute, set it on the extra hash of an update so it can be passed through to FMS for use in templates.

Discussion on ticket about implementing this:
https://github.com/mysociety/societyworks/issues/5472

### DIFF
--- a/perllib/Open311/Endpoint/Integration/UK/Dumfries.pm
+++ b/perllib/Open311/Endpoint/Integration/UK/Dumfries.pm
@@ -266,21 +266,44 @@ sub _get_inspection_updates_design {
             if ($mapping && $mapping->{extra_attributes} && $mapping->{extra_attributes}{latest_inspection_time}) {
                 my $latest_inspection = $self->_find_latest_inspection($report);
                 if ($latest_inspection) {
-                    my $inspection_attrs = $self->alloy->attributes_to_hash($latest_inspection);
-                    my $completion_time = $inspection_attrs->{attributes_tasksCompletionTime};
-
-                    if ($completion_time) {
-                        $completion_time = $completion_time->[0] if ref $completion_time eq 'ARRAY';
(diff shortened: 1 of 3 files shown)


perllib/Open311/Endpoint/Integration/UK/Dumfries.pm
```diff
@@ -283,6 +275,38 @@ sub _get_inspection_updates_design {
...
+
+    if ($completion_time) {
+        $completion_time = $completion_time->[0] if ref $completion_time eq 'ARRAY';
+        $update->{extras}{latest_inspection_time} = $completion_time;
+    } else {
+        $update->{extras}{latest_inspection_time} = 'NOT COMPLETE';
+    }
+
+    my $mapping = $self->config->{other_responsibility_mapping};
+    if (
```
davea, line 300: If `other_responsibility_mapping` isn't set in config this prints a bunch of warnings, maybe we should check for `$mapping` too?

perllib/Open311/Endpoint/Integration/UK/Dumfries.pm
```diff
@@ -283,6 +275,38 @@ sub _get_inspection_updates_design {
...
+        my $responsibility = $inspection_attrs->{ $mapping->{'highway'} }
+        || $inspection_attrs->{$mapping}{'streetlight'}
```
davea: This is inconsistent with the structure on the previous line, I think it should be this:

```suggestion
        || $inspection_attrs->{ $mapping->{'streetlight'} }
```

davea (changes requested):
A couple of small things in the code. However I've also not actually been able to get this to return any values even when running against the live API. So not quite sure what's going on.

davea (approved):

PR #514, opened 2026-05-15
[Dumfries] Add itemsTitle from asset layer to Enquiry

MorayMySoc:
Set the location field of the Alloy Enquiry to contain the title from the asset layer

https://github.com/mysociety/societyworks/issues/5473

### DIFF
--- a/perllib/Open311/Endpoint/Integration/UK/Dumfries.pm
+++ b/perllib/Open311/Endpoint/Integration/UK/Dumfries.pm
@@ -28,6 +28,22 @@ sub process_attributes {
     my ($self, $args) = @_;
 
     my $attributes = $self->SUPER::process_attributes($args);
+    #If there is an asset_resource_id will have cached the call
+    # for the item in set_parent_attribute
+    my $asset_attributes = $args->{attributes}->{asset_resource_id}
+      ? $self->api_cache->{ $args->{attributes}->{asset_resource_id} }
+      : '';
+    if (!($asset_attributes && $asset_attributes->{item})) {
+        $args->{attributes}->{asset_resource_id}
+          ? $self->logger->warn("Failed to fetch highway item " . $args->{attributes}->{asset_resource_id})
+          : $self->logger->warn("No asset_resource_id in report");
+    } else {
+        $asset_attributes = $self->alloy->attributes_to_hash($asset_attributes->{item});
+        push @$attributes, {
+            attributeCode => $self->config->{request_to_resource_attribute_manual_mapping}->{customer_enquiry_location_field},
+            value => $asset_attributes->{ 'attributes_itemsTitle' },
+        };
+    }
 
     my $contact_resource_id = $self->_find_or_create_contact($args);
     push @$attributes, {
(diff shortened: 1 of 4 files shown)


perllib/Open311/Endpoint/Integration/UK/Dumfries.pm
```diff
@@ -29,6 +29,21 @@ sub process_attributes {
 
     my $attributes = $self->SUPER::process_attributes($args);
 
+    my $road_attributes = $args->{attributes}->{asset_resource_id}
+      ? $self->alloy->api_call(call => "item/" . $args->{attributes}->{asset_resource_id})
```
dracos: We will have already called this in `set_parent_attribute`, so can we somehow store the call result made there so we don't have to call it again. Perhaps we could have an api_cache object on $self or similar?

perllib/Open311/Endpoint/Integration/UK/Dumfries.pm
```diff
@@ -29,6 +29,21 @@ sub process_attributes {
...
+      : '';
+    if (!($road_attributes && $road_attributes->{item})) {
```
dracos: For Street lights this will be a street light item, not a road; I can't tell from the name/discussion if that matters or not - my guess is not, and the street light title would be just as good to use.

perllib/Open311/Endpoint/Integration/UK/Dumfries.pm
```diff
@@ -29,6 +29,21 @@ sub process_attributes {
...
+        $args->{attributes}->{asset_resource_id}
+          ? $self->logger->warn("Failed to fetch highway item " . $args->{attributes}->{asset_resource_id})
+          : $self->logger->warn("No asset_resource_id in report");
+    } else {
+        $road_attributes = $self->alloy->attributes_to_hash($road_attributes->{item});
+        push @$attributes, {
+            attributeCode => $self->config->{request_to_resource_attribute_manual_mapping}->{customer_enquiry_location_field},
+            value => $road_attributes->{ $self->config->{request_to_resource_attribute_manual_mapping}->{highways_asset_layer_road_name} },
```
dracos: is this `attributes_itemsTitle`? I think we're okay hard-coding that because all objects have that.

dracos (changes requested):

perllib/Open311/Endpoint/Integration/UK/Dumfries.pm
```diff
@@ -28,6 +28,22 @@ sub process_attributes {
     my ($self, $args) = @_;
 
     my $attributes = $self->SUPER::process_attributes($args);
+    #If there is an asset_resource_id will have cached the call
+    # for the item in set_parent_attribute
+    my $asset_attributes = $args->{attributes}->{asset_resource_id}
```
MorayMySoc, line 33: If I've assigned this, I should probably be using $asset_attributes going forward...

perllib/Open311/Endpoint/Integration/UK/Dumfries.pm
```diff
@@ -28,6 +28,22 @@ sub process_attributes {
...
+      ? $self->api_cache->{ $args->{attributes}->{asset_resource_id} }
```
MorayMySoc, line 34: Ah, ok. Doesn't work here...

dracos (approved):
Thanks

PR #515, opened 2026-05-27
[Aurora] Handle 5K limit when getting filenames from Azure Storage Container.

neprune:
For https://mysocietysupport.freshdesk.com/a/tickets/7012.

Verfiied locally that the missing updates come through with this.

### DIFF
--- a/perllib/Integrations/Aurora.pm
+++ b/perllib/Integrations/Aurora.pm
@@ -358,17 +358,37 @@ Queries for filenames in the 'return path updates' Azure storage container.
 sub fetch_update_filenames {
     my ($self) = @_;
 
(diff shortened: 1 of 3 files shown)


dracos (approved):
As long as we're sure the filenames are in UTC :)
Could also bail out if marker == next_marker I guess, to prevent a loop there, but assume that shouldn't ever happen.

neprune:
> As long as we're sure the filenames are in UTC :) Could also bail out if marker == next_marker I guess, to prevent a loop there, but assume that shouldn't ever happen.

Whoops, yep, out they are currently on BST going by the timestamp recorded in the file vs the filename. This meant we also had an existing bug in `_skip_update_file` where we were an hour late to reading update . Although thankfully we are at least still populating the update time from the UTC timestamp in the update file ok. 

Have updated to use `Europe/London`.

PR #519, opened 2026-07-02
[Abavus] Allow multiple categories with same code.

dracos:
Similar in principle to how some others such as Confirm.

Should stop the cron email complaining about multiple entries.


perllib/Open311/Endpoint/Integration/Abavus.pm
```diff
@@ -224,6 +225,8 @@ sub post_service_request {
 
     my @address = split /\r?\n/, ($args->{attributes}{closest_address} || '');
 
+    my $code = $args->{service_code};
+    $code =~ s/_[0-9]$//;
```
MorayMySoc: Do we restrict all integrations to only 9 aliases for a category?

MorayMySoc (approved):
Only comment is that the ceiling on aliases looks like it could trip us up at some point if I've understood that correctly.

perllib/Open311/Endpoint/Integration/Abavus.pm
dracos: Camden Confirm does have 11, I guess, and Gloucestershire does not have more, but does use higher numbers due to category name changes, so yeah, it might as well handle more

PR #520, opened 2026-07-08
[Northumberland] Questions on categories map to Alloy attributes

MorayMySoc:
Where a category has extra questions in FMS, allow these answers to be mapped via configuration in 'request_to_resource_attribute_manual_mapping' to Alloy attributes.

https://mysocietysupport.freshdesk.com/a/tickets/7248

### DIFF
--- a/perllib/Open311/Endpoint/Integration/UK/NorthumberlandAlloy.pm
+++ b/perllib/Open311/Endpoint/Integration/UK/NorthumberlandAlloy.pm
@@ -47,6 +47,7 @@ In addition to the default new request processing, this function:
 * Gets category and group codes from the provided data.
 * Looks up the category via C<category_list_code> and C<category_title_attribute>, adding this item under the 'category' attribute specified in C<request_to_resource_attribute_manual_mapping>.
 * Looks up the category via C<group_list_code> and C<group_title_attribute>, adding this item the 'group' attribute specified in C<request_to_resource_attribute_manual_mapping>.
+* Looks up attributes for extra questions for specific categories in C<request_to_resource_attribute_manual_mapping> and maps answers to the Alloy attributes
 
 =cut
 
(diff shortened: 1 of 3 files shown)


perllib/Open311/Endpoint/Integration/UK/NorthumberlandAlloy.pm
```diff
@@ -61,6 +62,16 @@ sub process_attributes {
         value => [ $contact_resource_id ],
     };
 
+    if (my $mapping = $self->config->{request_to_resource_attribute_manual_mapping}{ $args->{service_code_alloy} }) {
+        for my $key ( keys %$mapping ) {
+            push @$attributes,
```
dracos, line 68: Do we want to add a `next` if args->attributes->{$key} isn't set or will it always be?

dracos (commented):
Looks good, just a question about if we want to skip missing values

chrismytton (approved):

perllib/Open311/Endpoint/Integration/UK/NorthumberlandAlloy.pm
```diff
@@ -61,6 +62,17 @@ sub process_attributes {
         value => [ $contact_resource_id ],
     };
 
+    if (my $mapping = $self->config->{request_to_resource_attribute_manual_mapping}{ $args->{service_code_alloy} }) {
+        for my $key ( keys %$mapping ) {
+            next unless $args->{attributes}->{$key};
+            push @$attributes,
+              {
+               attributeCode => $mapping->{$key},
+               value => [ $args->{attributes}->{$key} ],
```
dracos: For the entries this is for at present, this should be `value => $args->{attributes}->{$key},` as they're  string values.

dracos (changes requested):
Didn't realise this at the time!

dracos (approved):
Hopefully that should do it :)

m7xedhoward:
If using old style config, the questions could be skipped

[https://github.com/mysociety/open311-adapter/blob/master/perllib/Open311/Endpoint/Integration/AlloyV2.pm#L161](https://github.com/mysociety/open311-adapter/blob/master/perllib/Open311/Endpoint/Integration/AlloyV2.pm#L161)

dracos:
@m7xedhoward Thanks; the questions in this case are currently set up in the admin, so that doesn't matter here, but I agree it'd be good for both flows to handle question configuration if possible (and/or stop using the old-style config in time).

PR #521, opened 2026-07-13
[Dudley] Symology integration

dracos:
For https://github.com/mysociety/societyworks/issues/5597 and https://github.com/mysociety/societyworks/issues/5599

Need to document overloading of `account_id` for update ID.

Awaiting config to make it match more what they have, but I think this should cover

### DIFF
--- a/perllib/Open311/Endpoint/Integration/Symology.pm
+++ b/perllib/Open311/Endpoint/Integration/Symology.pm
@@ -22,6 +22,7 @@ with 'Role::Logger';
 
 use Integrations::Symology;
 use Open311::Endpoint::Service::Attribute;
+use Open311::Endpoint::Service::Request::ExtendedStatus;
 use Open311::Endpoint::Service::Request::Update::mySociety;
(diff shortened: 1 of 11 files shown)


perllib/Open311/Endpoint/Integration/Symology.pm
```diff
@@ -510,4 +511,49 @@ sub event_action_event_type {
...
+
+Fetches a list of requests from the backend, really for updates.
+
+=cut
+
+sub service_request_content {
+    '/open311/service_request_extended'
+}
+
+sub get_service_requests {
```
davea, line 524: Should this (and `service_request_content` above) live in `Dudley.pm` rather than here where it affects all Symologies?

davea (approved):
One scoping thing, but looks OK for staging at least :)

perllib/Open311/Endpoint/Integration/Symology.pm
dracos: I did wonder, but I guess any other Symology could also utilise this same method

PR #524, opened 2026-08-07
[CRT] Canals and river integration fetching

MorayMySoc:
Adds integration for Canals and River Trust using the Tech Spec

Posts FMS reports as Cases and Incidents.

Fetches Incidents to add as FMS reports.

Fetches updates for incidents (in case of status change)

https://github.com/mysociety/societyworks/issues/5587

(canals-river-trust for server side configuration)

### DIFF
--- a/perllib/Open311/Endpoint/Integration/Sugar.pm
+++ b/perllib/Open311/Endpoint/Integration/Sugar.pm
@@ -1,0 +1,92 @@
+=head1 NAME
+
+Open311::Endpoint::Integration::Sugar - An integration with the Sugar CRM system
+
+=head1 SYNOPSIS
+
+This integration lets us populate categories
+
+=cut
+
+package Open311::Endpoint::Integration::Sugar;
+
+use Moo;
+
+extends 'Open311::Endpoint';
+with 'Open311::Endpoint::Role::ConfigFile';
+
+=head2 jurisdiction_id
+
+Has the jurisdiction_id for matching reports to sugar
+
+=cut
+
+has jurisdiction_id => (
+    is => 'ro',
+);
+
+
+=head2 service_list
+
+This is a mapping of Sugar services to use for categories populating FMS.
+
+=cut
+
+has service_list => (
+    is => 'ro',
+);
+
+=head2 service_extra_data
+
+This is a mapping of Sugar attributes. Most categories
+will have an  asset_id and some have extra questions
+
+=cut
+
+has service_extra_data => (
+    is => 'ro',
+);
+
+
+=head2 service_class
+
+Subclasses can override this to provide their own custom Service class, e.g.
+if they want to have extra attributes on all services.
+
+=cut
+
+has service_class  => (
+    is => 'ro',
+    default => 'Open311::Endpoint::Service::UKCouncil',
+);
+
+sub services {
+    my $self = shift;
+
+    my @services = ();
+    for my $group (sort keys %{ $self->service_list }) {
+        my $categories = $self->service_list->{$group};
+        my $service_code;
+        (my $format_group = $group) =~ s/\s+/-/g;
+        for my $category (sort keys %{ $categories }) {
+            ($service_code = $category) =~ s/\s+/_/g;
+            my %service = (
+                service_name => $category . ' (CRT: ' . $group . ')',
+                description => $category,
+                service_code => $format_group . '_' . $service_code,
+                group => $group,
+            );
+            my $o311_service = $self->service_class->new(%service);
+            my $extra = $self->service_list->{$group}{$category}{service_extra_data};
+            foreach (@$extra) {
+                $_->{datatype} = 'singlevaluelist';
+                push @{$o311_service->attributes}, Open311::Endpoint::Service::Attribute->new(%$_);
+            }
+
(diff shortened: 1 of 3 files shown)


MorayMySoc:
I've used a '-' to mark spaces in group names when creating the service code - but this doesn't work with groups like 'Bridge - lift / swing' that already contain a hyphen

t/open311/endpoint/canals.t
```diff
@@ -0,0 +1,28 @@
...
+BEGIN { $ENV{TEST_MODE} = 1; }
+
+my $canals_endpoint = Open311::Endpoint::Integration::Sugar::Dummy->new(
+    jurisdiction_id => 'canals_sugar',
+    config_file => path(__FILE__)->sibling("canals.yml")->stringify,
+    );
+
+subtest "GET Service List" => sub {
+    my $res = $canals_endpoint->run_test_request( GET => '/services.xml' );
+    ok $res->is_success, 'xml success';
```
dracos, line 67: I'd probably check the contents look okay here :)

perllib/Open311/Endpoint/Integration/Sugar.pm
```diff
@@ -25,6 +28,72 @@ has jurisdiction_id => (
...
+=cut
+
+has integration_class => (
+    is => 'ro',
+    default => 'Integrations::Rest'
+);
+
+=head2 sugar
+
+Instantiate the configuartion as sugar.
```
dracos: typo in configuration

perllib/Open311/Endpoint/Integration/Sugar.pm
```diff
@@ -25,6 +28,72 @@ has jurisdiction_id => (
...
+
+has client_id => (
+    is => 'ro',
+);
+
+=head2 access_token
+
+API calls must pass an access_token which is fetched using the username and password.
+It lasts an hour which should be enough for the lifetime of open311 requests so
+no need to save the refresh token
```
dracos, line 94: I wonder if they'd like us to save the tokens and use/refresh them, rather than request new access tokens over and over. We don't store them, as you say, so guess it doesn't matter

perllib/Open311/Endpoint/Integration/Sugar.pm
```diff
@@ -192,4 +201,135 @@ sub _do_login {
...
+    my $response = $self->rest->api_call(
+                                          call => $self->api_calls->{contact},
+                                          headers => $self->headers,
+                                          body => {
+                                                   first_name => $args{first_name},
+                                                   last_name => $args{last_name},
+                                                   email1 => $args{email},
+                                                  },
+                                        );
+    $self->rest->return_json_error(0);
```
dracos, line 344: If this is a per-call flag, not for the whole Rest class, is it better done as a new flag in `api_call`? Or do we not know at this point what errors we'll be handling.

perllib/Open311/Endpoint/Integration/Sugar.pm
```diff
@@ -332,4 +347,116 @@ sub _get_user {
...
+        $args->{start_date} = DateTime->now->set_time_zone('Europe/London') - DateTime::Duration->new( days => 1 );
+    };
+
+    if (!$args->{end_date}) {
+        $args->{end_date} = DateTime->now->set_time_zone('Europe/London');
+    }
+
+    $self->_do_login;
+    my $filter = _generate_filter(
+                  '[modified_user_id][$not_equals]=' . 'b7c78032-144f-11eb-b85e-00237dd1fdde',
```
dracos: Should this value be in config?

perllib/Open311/Endpoint/Integration/Sugar.pm
MorayMySoc: It is under discussion whether this process is going to happen at all now, so I may just leave this as it is for the time being

t/open311/endpoint/canals.t
MorayMySoc: I've updated ```canals-river-trust``` config with the new layout (option 1). We are still waiting for the categories to be confirmed though

perllib/Open311/Endpoint/Integration/Sugar.pm
```diff
@@ -0,0 +1,465 @@
...
+Owner id of incidents created by FMS
+
+=cut
+
+has crm_user_id => (
+    is => 'ro',
+);
+
+has headers => (
+                is => 'lazy',
```
nephila-nacrea: There appears to be too much indentation here (and some other places further down). I think for Perl files we generally default to 4 spaces per level?

perllib/Open311/Endpoint/Integration/Sugar.pm
```diff
@@ -0,0 +1,465 @@
...
+};
+
+sub _create_incident {
+    my ($self, $args) = @_;
+
+    my %defaults = (
+                    resolution => 'Accepted',
+                    type => 'Administration',
+                    status => 'New',
+                    resolution => 'Accepted',
```
nephila-nacrea: 'resolution' appears twice.

dracos (approved):
Thanks :)

